PR #194, opened 2015-01-15
Highlight port and connections on hover.

egasimus:
Testing the waters. IMHO this makes the patchbay feel more responsive, but let me know if it clashes with existing behavior.

I'm also thinking of implementing the following behavior: if you're dragging a wire from e.g. an audio output, then all ports that are not audio inputs are darkened so you can easily see the possible destinations for dropping that wire. Would you be opposed to me refactoring the themes first, though?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 # Temp files
 *~
 *.bak
+*.swp
+*.swo
 
 # Binary files
 *.7z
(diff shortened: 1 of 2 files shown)


falkTX:
Not sure how I feel about this, I'll have to give it a try.

Darkening the incompatible connections sounds good.
There's some pending work I need to do to support alternate ports (MIDI-OSC and Audio-CV), so you should wait until I do that first.

I'm also not sure about reworking the themes right now.
I don't want to delay the beta4 release and this seems like something that may take a while to do and test.
You're free to do it, just don't expect it to be pulled-in so soon.

falkTX:
I must say I like this new option.
I made it work for client "groups" as well.
It makes carla patchbay require less clicks, which is nice.

I added an option for this in case users prefer the old behaviour.

PR #198, opened 2015-01-26
new midisequencer UI

rhetr:
updated UI for midisequencer with note add/remove and some other stuff.

keys:
F - insert mode
D - velocity mode
A - select/deselect all
Del/Backspace - delete selected

### DIFF
--- a/source/native-plugins/resources/midiseq-ui
+++ b/source/native-plugins/resources/midiseq-ui
@@ -3,7 +3,7 @@
 
 # A piano roll viewer/editor
 # Copyright (C) 2012-2015 Filipe Coelho <falktx@falktx.com>
-# Copyright (C) 2014-2015 tatch
+# Copyright (C) 2014-2015 Perry Nguyen
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


falkTX:
Nice, thanks very much!

Should I remove the initial notes?
If you think the add/removal of notes works well enough, I can make it save&load state to make this plugin really work.

btw, I'd love if it had a menu or toolbar.
You can create a qt *.ui file for that like I do with the main Carla stuff.

rhetr:
note add/removal works pretty well except I've noticed with the initial notes that I sometimes get these errors upon removal:

    MidiPattern::removeRaw(419803, 0x7f79ebdb0c60, 3) - unable to find event to remove
    MidiPattern::removeRaw(369115, 0x7f79ebdb0c60, 3) - unable to find event to remove

I think it most likely has something to do with the way I convert note information from time to beats and vice versa in `handleMidiEvent()` and `sendMsg`. If that's the case then that problem may only arise when notes are inserted via something other than the GUI, in which case I think that could be fine for now, unless you know of a better way to do it off the top of your head.

falkTX:
Well, the time of events should be in PPQ not frames.
But this is not something I'm very good at, so we'll have to keep trying until it works I think.
Do you think you can handle time being PPQ for midi messages?

afaik with PPQ notes can be aligned properly even after BPM changes.

You'll probably need to know the "TicksPerBeat", but not sure.
Whatever data is needed to make a proper time implementation let me know.

rhetr:
I'm not familiar with PPQ but the UI's internal notes are kept in terms of beats so that sounds like a good idea. Does PPQ equal TicksPerBeat or is that only when you are in 4/4 time?

falkTX:
Well, it can be any format, PPQ just seem to be the most common one (mostly because of VST).

Timing in BBT can work too.
It would just need more values.

How would you handle time signature changes in this case?

rhetr:
I'm already using JACK time signature information to convert messages being sent between the UI and the backend. Right now time is being converted into quarter notes (ie beats) like this:

    start = time_ / 60. * self.fTransportInfo["bpm"]
    length = (time - time_) / 60. * self.fTransportInfo["bpm"] / 4. / self.fTransportInfo["sigNum"] * 

self.fTransportInfo["sigDenom"]
(we should multiply `start` by ` 1 / 4. * self.fTransportInfo["sigNum"]` too so it properly incorporates the time signature.)

I think basically PPQ is 

    time_ / 60. * self.fTransportInfo["bpm"] == PPQ / total_number_of_pulses_in_quarter_note

so you just substitute that for PPQ and it should work.

If the heart of the problem lies in the way python manipulates floats I may have to change the way notes are being stored internally in the UI but I think switching to PPQ as a first step is a good idea.

falkTX:
ok, I looked a bit at the juce code that handles midi files and it uses "TicksPerQuarterNote".
the "time" value we get will use ticks.

we can define any tick resolution we want, but better to choose one that is nice for the UI.

(let me write this to make this easier to understand)
let's say we're playing in 4/4 signature, that means each phrase/part has 4 bars, and each bar has 4 beats.
the question is how many ticks we want in a beat. 24? 48? 96? more?

tell me which value you think it's more adequate.
and let me know if I got anything wrong...

rhetr:
I think that sounds right. Hm, I have no idea how many ticks we want. it looks like they use [480 ticks](https://cycling74.com/forums/topic/midi-beat-clock/) per beat internally for max/msp but that sounds like overkill for us. How about we try 48 for now? that sounds reasonable enough I think.

falkTX:
new code is in, now events are handled in ticks.
I went with 48 ticks per beat.

this allows the playing speed to change according to bpm.
there is no fancy sync though, for example with external apps that have a tempo ramp.
JACK transport doesn't support this, so less work for us.

the UI code needs an update now, I'll leave that you.

btw, I noticed that if there's note-on and note-off at the same tick pos, the note-off may not be handled before the note-on.
I'll fix this soon.

PR #206, opened 2015-02-08
use 48 ppq for midiseq-ui

rhetr:
I also included code for moving the playhead according to the transport. Right now since self.fTransportInfo only updates frames it only works correctly with 120bpm since I just hardcoded a value that worked. I think it would work with any bpm if we could pass ticks instead (it's in self.fTransportInfo already but doesn't seem to update).


falkTX:
The ticks info doesn't update unless a jack-transport master client is running.
This is wrong though, as we use a custom ticksPerBeat so we shouldn't care about jack-transport for this.

I'll do a commit to send the same info the backend uses.

PR #254, opened 2015-05-16
Add "--version" command line option.

egasimus:
Prints version info and exits. Might be useful.
`--help` is up next, and then perhaps `--no-gui`. But I'll try to DRY up the four executables before that -- no use repeating the same code four times. Looking forward to any comments.


falkTX:
This pull request has 4/6 commits that are useless.
I don't like the way you made them anyway.

I know carla needs to handle command-line arguments like --version, --help, etc.
But with this you're just adding more to the problem.
We need to centralize the options so all apps (carla, carla-rack, etc) get them.

Can you simply report a bug instead?
I'll look into implementing CLI args after I finish some pending changes.

egasimus:
Your response is actually kind of rude :(

>This pull request has 4/6 commits that are useless.

...because you never merged #194.
I guess I should've run `git rebase` before submitting or something.
Sorry, still fear that command a little...

>I know carla needs to handle command-line arguments like --version, --help, etc.
>But with this you're just adding more to the problem.

...you seem to be saying that I am exacerbating the problem by _solving it_? (Try as I might, I could not understand these sentences in another way.) Yes, #254 is solving it in a somewhat suboptimal way -- which is why I said `But I'll try to DRY up the four executables before that -- no use repeating the same code four times` and submitted #255 immediately afterwards. Which, it seems, you've only rejected because you had already rejected this #254...

> I don't like the way you made them anyway. ... I'll take your changes as base for my code, thanks.

...oh, well, that's, erm, reassuring. :D

Nevermind. I respect your work and vision (which is also why I was trying to contribute; which I guess I might not be doing as enthusiastically in the future). So take your time. I tried to save you some but it seems I've only wasted mine :/

falkTX:
sorry I first saw #254 and didn't know about #255 when I wrote that.
so my answer is only relevant to the first pull request (which did not have the carla_main.py stuff).

I did implement your "Highlight port and connections on hover" request, in a little different way.
There's a new option in the canvas settings for this. Did you miss it?
I also pushed your "Ignore vim .swp/.swo tempfiles" bit.
Am I missing something from #194 ?

egasimus:
What happened with #194 was since you only merged that vim commit but reimplemented the feature commit yourself, and left the PR unmerged, my fork ended up in a state where I couldn't merge cleanly from upstream -- hence the two duplicated and the two useless commits in this PR. It doesn't look like it would be difficult to clean up if I had a little more git-fu, but it did sort of show things in a bad light if you know what I mean.

Nevermind that -- why not rather talk about #255? Is there anything you'd like to see improved if I wanted to get that stuff merged after all or do I just leave everything to you? (As initially implied, the actual reason I did everything was because I need a headless plugin host and Carla's the best there is -- so I wanted to try my luck at making the GUI part optional.)

falkTX:
When something like that happens (upstream implements a different version from my commit)
I usually delete the repo and start again.
I had to do this a few times already.

Anyway, about #255 I'll implement something based on your code soon.
Thanks.

PS: the carla-single script can handle headless setups, please try it

egasimus:
:+1: Alright, thanks for your advice :)

falkTX:
I looked into implementing something like your carla_main.py but the small variations make it a big mess that cannot be handled nicely.

People have been asking for a no-gui mode so that's what I'm going to do first.

PR #255, opened 2015-05-16
Single entry point

egasimus:
The successor to https://github.com/falkTX/Carla/pull/254.

* Combines the duplicated code from `carla`, `carla-rack`, `carla-patchbay` and `carla-control` into the new `carla_main.py`.
* Adds usage info -- printed with `carla --usage` or `carla --help` (the difference is that the latter also prints the output of `carla --version`).
* Fixes the following bug which was (probably) setting the wrong app name:
```
if arg.startswith("--with-appname="):
    initName = os.path.basename(arg.replace("--with-initname=", ""))
```


falkTX:
Same as #254 I'll handle this myself later on (please do a bug report).
I'll take your changes as base for my code, thanks.

PR #307, opened 2015-10-07
Fix installation of themes

kasbah:
My best guess as to what it should be.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -502,7 +502,7 @@ endif
 ifeq ($(HAVE_QT),true)
 	# Install theme
 	install -m 644 \
-		bin/styles/* \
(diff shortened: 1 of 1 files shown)


falkTX:
bin/styles should be automatically created when the styles are built.
is it empty for you?

kasbah:
It is yeah, even after `make`. Just doing another pull request with build instructions. Maybe you could point me to what I am missing?

falkTX:
You're probably missing the headers for Qt.
The theme is c++ code, not python.

kasbah:
Well, I am not looking to package anyway so I was barking up the wrong tree with this.

PR #308, opened 2015-10-07
Add build instructions to README

kasbah:


falkTX:
Build instructions are the INSTALL file.
And you don't need to install to run carla, you can just do `./source/carla`.

Also I don't recommend building with PyQt5, as it will miss the VST plugin on Linux.

PR #311, opened 2015-10-11
Add a hide file browser menu option (fixes #199)

### LINKED ISSUE #199: include option to hide file browser
I have never used the file browser and would like to be able to hide it.

kasbah:
This defaults to the panel still showing up on newly updated versions.

### DIFF
--- a/resources/ui/carla_host.ui
+++ b/resources/ui/carla_host.ui
@@ -226,6 +226,7 @@
     <addaction name="act_settings_show_toolbar"/>
     <addaction name="act_settings_show_meters"/>
     <addaction name="act_settings_show_keyboard"/>
+    <addaction name="act_settings_show_side_panel"/>
     <addaction name="separator"/>
     <addaction name="act_settings_configure"/>
    </widget>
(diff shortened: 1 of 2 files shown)


falkTX:
the left side is not just for the filebrowser, there will be other stuff there later.

one thing to note, the line:
```
settings.setValue("ShowFileBrowser", self.ui.dockWidget.isVisible())
```
should be:
```
settings.setValue("ShowFileBrowser", self.ui.dockWidget.isEnabled())
```
because carla's window could be hidden by the OS (ie, minimized) the visible check can sometimes return false.

kasbah:
Yes, I did see that mentioned in issue #199. I just feel calling it something else in the menu right now (while the new stuff isn't there yet) might be a bit confusing.

kasbah:
I changed it to `isEnabled` by the way. If you have a suggestion for a better name in the code and menu then I'll make that change too.

falkTX:
I'd call it SidePanel.

kasbah:
Ok, should I change the menu option to "Show Side Panel" too?

falkTX:
yes please.

kasbah:
I have made the changes.

falkTX:
Looks good, thanks.

PR #314, opened 2015-10-11
Detect PyQt from directories in PATH

joelmo:
Hi, do you think this change is ok?

I am on a system that doesn't use the FSH.


falkTX:
Those variables have "?=" meaning if the environment variable for them is defined it will replace that value.
so you can just use something like this:
```
export PYUIC4=/opt/something/pyuic4
make
```

joelmo:
Aha ok then I will use something like that

PR #355, opened 2016-01-30
Removed shebang from modules not meant to be executed as script

susnux:
Shebangs should be only used if the file is meant to be executed as a script.
And if so it should have the executable-flag set (because if not it is even with a shebang not executable).

Some packaging utilities (like rpm with rpmlint) will warn if there are non-executable files with shebang.


falkTX:
Why is this important to you?
I actually use some of those files as scripts, by running them directly to test the features of the file.
Example: https://github.com/susnux/Carla/blob/master/source/carla_settings.py#L847

Having the shebang is also better because it lets the IDE know the real filetype.
A file ending with *.py is not enough information as there are different versions of python.
The shebang can be used to report the file was coded for python3.

So, sorry but I don't agree with the changes.

susnux:
Ok I can understand you.
I noticed that because RPMLints "non-executable-script" (scripts contains shebang but not executable).
But ok, I will ignore the warnings.

PR #373, opened 2016-03-09
Fix build on GCC 5.3.0, where isnan ends up being undeclared otherwise.

jypma:
@falkTX without this I'm getting
```
Compiling zynaddsubfx-src.cpp
In file included from zynaddsubfx/DSP/FFTwrapper.cpp:27:0,
                 from zynaddsubfx-src.cpp:74:
zynaddsubfx/DSP/FFTwrapper.h: In instantiation of ‘std::complex<_Tp> FFTpolar(const _Tp&, const _Tp&) [with _Tp = float]’:
zynaddsubfx/Params/PADnoteParameters.cpp:872:68:   required from here
zynaddsubfx/DSP/FFTwrapper.h:61:18: error: ‘isnan’ was not declared in this scope, and no declarations were found by argument-dependent lookup at the point of instantiation [-fpermissive]
         if (isnan(__x))
                  ^
In file included from /usr/include/c++/5.3.0/random:38:0,
                 from /usr/include/c++/5.3.0/bits/stl_algo.h:66,
                 from /usr/include/c++/5.3.0/algorithm:62,
                 from zynaddsubfx/rtosc/cpp/midimapper.cpp:4,
                 from zynaddsubfx-src.cpp:58:
/usr/include/c++/5.3.0/cmath:641:5: note: ‘template<class _Tp> constexpr typename __gnu_cxx::__enable_if<std::__is_integer<_Tp>::__value, bool>::__type std::isnan(_Tp)’ declared here, later in the translation unit
     isnan(_Tp __x)
     ^
```
on the latest master.

### DIFF
--- a/source/native-plugins/zynaddsubfx-src.cpp
+++ b/source/native-plugins/zynaddsubfx-src.cpp
@@ -25,6 +25,8 @@
 using std::isnan;
 #else
 # include <err.h>
+#include <cmath>
+using std::isnan;
 #endif
 
 #define PLUGINVERSION


falkTX:
I appreciate the fix, but please send it to the zynaddsubfx project instead.
Once that is merged, I'll update carla's copy and it will build again.

But anyway, this has already been fixed in current carla code.
See https://github.com/falkTX/Carla/blob/master/source/native-plugins/zynaddsubfx/DSP/FFTwrapper.h#L52

Your carla fork is old :)

jypma:
Hmm interesting :) I'll double-check, thanks for the info.

PR #384, opened 2016-03-28
Fix #380: Limit stack allocation during base64 encoding

### LINKED ISSUE #380: Stack overflow in carla-bridge-win32 when transport state changes
I haven't yet fully determined what the root cause is, but Kontakt 5 inside carla-bridge-win32 inside carla-lv2 in ardour (pff...) regularly crashes with a "stack overflow in carla-bridge-win32" when performing an action in ardour, or just letting it play for a moment. Not all 32-bit plugins do this, nor does it happen when running Kontakt 5 inside carla-bidge-win32 inside carla standalone.

@falkTX  I'm not really the expert at deciphering 32-bit back traces but does this indeed indicate the stack overflow is within carla? Is there a way to get function names into the backtrace?
```
Unhandled exception: stack overflow in 32-bit code (0x004126f3).
Register dump:
 CS:0023 SS:002b DS:002b ES:002b FS:0063 GS:006b
 EIP:004126f3 ESP:008df5d4 EBP:0095f6f8 EFLAGS:00210202(  R- --  I   - - - )
 EAX:00003000 EBX:00000056 ECX:007615e0 EDX:00000004
 ESI:0017fff5 EDI:09761017
Stack dump:
0x008df5d4:  00180000 008df5e0 0046a548 008df5f8
[...]
0x008df624:  686f5075 35783738 46487a4e 62725939
Backtrace:
=>0 0x004126f3 in carla-bridge-win32 (+0x126f3) (0x0095f6f8)
  1 0x0046df43 in carla-bridge-win32 (+0x6df42) (0x0095fc38)
  2 0x006ef114 in carla-bridge-win32 (+0x2ef113) (0x0095fd38)
  3 0x004013e2 in carla-bridge-win32 (+0x13e1) (0x0095fe10)
  4 0x7b860a7c call_process_entry+0xb() in kernel32 (0x0095fe28)
  5 0x7b861d38 in kernel32 (+0x41d37) (0x0095fe68)
[...]
  15 0xf756c937 wine_init+0x2c6() in libwine.so.1 (0xffde8848)
  16 0x7bf00f22 main+0x81() in <wine-loader> (0xffde8ca8)
  17 0xf73a8527 __libc_start_main+0xf6() in libc.so.6 (0x00000000)
0x004126f3: orl	$0,0x0(%e
```
Full backtrace [here](http://pastebin.com/raw/C6BAJKwZ).

jypma:
Some VST plugins want to save a large binary chunk of state. For example,
Drumlab inside Kontakt 5 saves 1.5MB of data. We shouldn't allocate that
much on the stack.

### DIFF
--- a/source/utils/CarlaString.hpp
+++ b/source/utils/CarlaString.hpp
@@ -20,6 +20,7 @@
 
 #include "CarlaJuceUtils.hpp"
 #include "CarlaMathUtils.hpp"
+#include <algorithm>
 
 // -----------------------------------------------------------------------
 // CarlaString class
@@ -585,7 +586,7 @@ public:
             "abcdefghijklmnopqrstuvwxyz"
             "0123456789+/";
 
-        const std::size_t kTmpBufSize = carla_nextPowerOf2(static_cast<uint32_t>(dataSize/3));
+        const std::size_t kTmpBufSize = std::min(carla_nextPowerOf2(static_cast<uint32_t>(dataSize/3)), 65536U);
 
         const uchar* bytesToEncode((const uchar*)data);
 


jypma:
This resolves the issue for me, now saves and loads Kontakt 5 with Drumlab correctly.

falkTX:
Change the `std::size_t(65536)` to `65536U` and I'll apply it :)

jypma:
Doesn't it need `uint32_t` as well, since on 64-bit an `U` literal is 64-bit? Or will that down-cast just fine?

falkTX:
on 32bits sizeof(size_t) == sizeof(uint32_t), so I don't think it needs anything else.

jypma:
sure, but doesn't the same code get compiled for the 64-bit bridge as well?

falkTX:
yes, and?
65536U is just so that std::min compares unsigned to another unsigned.
the returned value is converted to a size_t so there's no precision loss at any point.

jypma:
Thanks for the explanation, that's exactly what I was wondering about. I've removed the `size_t` cast now.

falkTX:
np, thanks for the fix!

PR #401, opened 2016-05-24
[fix][PyQt5.6]: launch errors; add plugin errors

pchome:
http://pyqt.sourceforge.net/Docs/PyQt5/incompatibilities.html


falkTX:
Seems good to me, thanks.

PR #475, opened 2017-02-28
Add option to allow disabling stripping of binaries

### LINKED ISSUE #474: Binaries should not be stripped during build
The current Makefiles strip all binaries during build when building on Linux. This shouldn't be the case. Stripping should happen either by the package manager or if it really has to be definied in the Makefile a separate `install-strip` target should be used.

See https://www.gnu.org/prep/standards/html_node/Standard-Targets.html for more info

Creating this issue because this causes a QA warning when building on Gentoo:
```
QA Notice: Pre-stripped files found:
<list of files>
```

simonvanderveldt:
Fixes #474

### DIFF
--- a/source/Makefile.mk
+++ b/source/Makefile.mk
@@ -72,7 +72,10 @@ ifeq ($(MACOS),true)
 LINK_OPTS  = -fdata-sections -ffunction-sections -Wl,-dead_strip -Wl,-dead_strip_dylibs
 else
 # Common linker flags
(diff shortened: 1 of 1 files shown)


falkTX:
hmm I didn't meant that way, although this can work too.
there should be no need to define default variables.

something like:
```
ifneq ($(SKIP_STRIPPING),true)
```

Also, don't change the raspberry options since you're likely not going to use `make RASPPI=true`

simonvanderveldt:
@falkTX Should be correct now :)

falkTX:
Alright, thanks!

This closes #474 too.

PR #519, opened 2017-09-26
Fix a couple of parallel build problems

atsampson:
I build carla with make -j8; these patches fix a couple of dependency problems that cause occasional parallel make failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,13 +112,13 @@ $(MODULEDIR)/jackbridge.%.a: .FORCE
 $(MODULEDIR)/native-plugins.a: .FORCE
 	@$(MAKE) -C source/native-plugins
 
-$(MODULEDIR)/theme.a: .FORCE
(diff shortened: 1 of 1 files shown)


falkTX:
I haven't encountered these myself, but they seem okay to me. merging now, thank you.

atsampson:
You undid these changes again in 8e5c2e292e229d6e4dbc28c64e5e3a55f777ed60 ("Move libjack into its own dir") - was that deliberate?

falkTX:
I undid part of it that was not needed.
only libjack needs "libs", interposer no longer needs it.

theme.qt4 and theme.qt4 cannot depend on theme itself, as that creates recursive dependency.
theme.qt* => theme => libs => theme.qt*

atsampson:
Aha, and you've also removed the theme.a rule, so make now can't enter source/theme twice. Looks good - thanks very much!

PR #547, opened 2017-12-01
import json in carla_host.py

jkroll20:
When saving I get this log entry:

```
Traceback (most recent call last):
  File "/usr/share/carla/carla_host.py", line 659, in slot_fileSave
    return self.saveProjectNow()
  File "/usr/share/carla/carla_host.py", line 594, in saveProjectNow
    json.dump({
NameError: name 'json' is not defined
```

Seems the import was missing.


falkTX:
Yeah, I see this can happen if rdflib is missing.
Thanks for the heads up.

PR #551, opened 2017-12-10
Simplify finding Qt5 compilers

dfoxfranke:
pkg-config exposes the variable `host_bins` for finding moc and friends, so the OS-specific breakdown isn't neaded.

### DIFF
--- a/source/Makefile.mk
+++ b/source/Makefile.mk
@@ -260,33 +260,10 @@ endif
 endif
 
 ifeq ($(HAVE_QT5),true)
-QT5_LIBDIR = $(shell pkg-config --variable=libdir Qt5Core)
-ifeq ($(BSD),true)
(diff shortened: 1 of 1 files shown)


falkTX:
This seems very nice.
Some time ago I tried to find if there was such an option, but only Qt4 provided a direct variable.

In which systems have you tested this in?
I have Carla setup to build currently for:
- Ubuntu 10.04 (for the generic linux binaries)
- Ubuntu 14.04
- Ubuntu 16.04
- ArchLinux
- FreeBSD
- MacOS
- HaikuOS

I will make sure this change works on all of them before accepting the PR.

dfoxfranke:
I've tested it (only) on NixOS, where the build breaks without this because it's radically non-FHS-compliant so every other way of locating moc fails.

PR #577, opened 2018-02-20
Delete bin/carla.lv2/styles (broken link)

trebmuh:
It looks like bin/carla.lv2/styles is a broken link.


trebmuh:
Sounds like a bad idea since the link looks to be broken "as is" but looks to be "unbroken" (ie: pointing to something relevant) while compiling.
I'm closing then.

falkTX:
The link is there for when running carla without installing.
Could be made nicer to create the link during build, but it doesn't hurt anyone afaik.

PR #584, opened 2018-03-01
properly handle mmap() errors

Jajcus:
Carla would ignore mmap() error in CarlaShmUtils.hpp and later crash on
0xffffffffffffffff pointer dereference.

The mmap() error could happen when locked memory ulimit is set too low. This causes Carla crashes for me when trying to load the Calf Organ plug-in on a badly configured system.


falkTX:
Nice catch, I was not aware mmap could return -1 for error.
I see some code there that I'd like to modify, so I'll import your PR manually.
Thanks!

PR #585, opened 2018-03-01
add semitones parameter to midi-transpose plugin

InTheMorning:
Works well both as internal and lv2 plugin.
Only issue is that the internal plugin still appears to only have 1 parameter from the "add plugin" window.
image

### DIFF
--- a/source/native-plugins/midi-transpose.c
+++ b/source/native-plugins/midi-transpose.c
@@ -26,7 +26,6 @@ typedef struct {
     const NativeHostDescriptor* host;
     int octaves;
     int semitones;
-    bool invert;
 } MidiTransposeHandle;
 
 // -----------------------------------------------------------------------
@@ -101,11 +100,13 @@ static float miditranspose_get_parameter_value(NativePluginHandle handle, uint32
     switch (index)
     {
         case 0:
-        return (float)handlePtr->octaves;
+            return (float)handlePtr->octaves;
+            break;
         case 1:
-        return (float)handlePtr->semitones;
+            return (float)handlePtr->semitones;
+            break;
         default:
-        return 0.0f;
+            return 0.0f;
     }
 }
 
@@ -114,11 +115,13 @@ static void miditranspose_set_parameter_value(NativePluginHandle handle, uint32_
   switch (index)
   {
       case 0:
-      handlePtr->octaves = (int)value;
+          handlePtr->octaves = (int)value;
+          break;
       case 1:
-      handlePtr->semitones = (int)value;
+          handlePtr->semitones = (int)value;
+          break;
       default:
-      return;
+          return;
   }
 
 }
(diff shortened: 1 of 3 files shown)


falkTX:
it's normal for the internal parameter count to not update, those are cached and only rescanned when the plugin count changes. the internal plugins were never designed to change much.
there's a small API_VERSION on the carla_database.py which can be bumped to force a refresh.

source/native-plugins/midi-transpose.c
```diff
@@ -50,57 +52,85 @@ static void miditranspose_cleanup(NativePluginHandle handle)
...
-    param.ranges.stepSmall = 1.0f;
-    param.ranges.stepLarge = 1.0f;
-    param.scalePointCount = 0;
-    param.scalePoints     = NULL;
-
-    return &param;
-
-    // unused
-    (void)handle;
+    static NativeParameter param[] =
```
InTheMorning: I wasn't sure if I should use a switch-case here, so I made 'param' an array.  Happy to know if this is good or bad.

falkTX: better to just see what the other plugins do and just copy them.
no shame in copy&pasting :)

InTheMorning: Haha yes!  For some reason this only occurred to me after the fact, but I left it as-is, because it compiled and ran as intended, and on the off-chance that this might be a better way, I wanted to show you :-)

source/native-plugins/midi-transpose.c
```diff
@@ -50,57 +52,85 @@ static void miditranspose_cleanup(NativePluginHandle handle)
...
+    return 2;
 
     // unused
     (void)handle;
 }
 
 static const NativeParameter* miditranspose_get_parameter_info(NativePluginHandle handle, uint32_t index)
 {
-    if (index != 0)
+    if (index >= miditranspose_get_parameter_count(handle))
```
falkTX: please change this to hardcoded `2`. we know this value in advance.

source/native-plugins/midi-transpose.c
```diff
@@ -50,57 +52,85 @@ static void miditranspose_cleanup(NativePluginHandle handle)
...
 {
-    if (index != 0)
-        return 0.0f;
-
-    return (float)handlePtr->octaves;
+    switch (index)
+    {
+        case 0:
+            return (float)handlePtr->octaves;
+            break;
```
falkTX: please remove this `break` it does nothing here as the return completes the execution of the function.
same with the one below it.

falkTX:
Looks great :)
Just do the 2 small changes I requested, let me know when done, and I'll merge the PR.

InTheMorning:
done :)

source/native-plugins/midi-transpose.c
```diff
@@ -116,10 +114,8 @@ static void miditranspose_set_parameter_value(NativePluginHandle handle, uint32_
     {
         case 0:
             handlePtr->octaves = (int)value;
-            break;
```
falkTX: hmm this is wrong now. you need the break here as it does not have a return, continue or other function stopper.
without the break, the cases fall through each other.

falkTX:
one small thing was overdone ;)

Issue #621, opened 2018-04-22
Feature request: MIDI Channel Filter but also with "Off" event-out


milkmiruku:
Two outputs, one for On, the other for Off, so, with instruments on different channels, one could build a chain of MIDI effects with insert control.

Extra feature potential: three outputs, the third a Thru, which passes everything so one could do a form of [half-normal](https://www.misterpatchbay.com/normaling/normal.html) style patching.

falkTX:
Not sure if I understood this request.
Feel free to make a PR for it, of course :)

milkmiruku:
Thanks for the merge!

Maybe I missed something, but the internal version has yourself as the Maker;
![rose_2018-06-28-10 18 57](https://user-images.githubusercontent.com/108225/42025984-df328c08-7abd-11e8-8d00-e62c5614a719.png)

falkTX:
The plugin meta-data is cached, so you get old data there.

here is my window:

![screenshot_20180628_113149](https://user-images.githubusercontent.com/1334853/42026223-da3f6f00-7ac6-11e8-9296-e8a573bb547f.png)

milkmiruku:
Grand, cheers :)

PR #622, opened 2018-04-25
MIDI Channel A/B

milkmiruku:
Added a new Carla internal plugin that sends each MIDI channel to either the A or B output (#621)


source/native-plugins/midi-channel-ab.c
```diff
@@ -0,0 +1,217 @@
...
+{
+    MidiGainHandle* const handle = (MidiGainHandle*)malloc(sizeof(MidiGainHandle));
+
+    if (handle == NULL)
+        return NULL;
+
+    handle->host = host;
+
+    for (int i=MAX_MIDI_CHANNELS; --i>=0;)
+        handle->channels[i] = 0.0f;
```
falkTX: This is not right, the channels are booleans, not floats.
And to zero memory we can use `memset(ptr, 0, full-size)` directly, instead of going through a list.

source/native-plugins/midi-channel-ab.c
```diff
@@ -0,0 +1,217 @@
...
+{
+    return MAX_MIDI_CHANNELS;
+
+    // unused
+    (void)handle;
+}
+
+static const NativeParameter* midichanab_get_parameter_info(NativePluginHandle handle, uint32_t index)
+{
+    if (index > MAX_MIDI_CHANNELS)
```
falkTX: This needs to be >=

source/native-plugins/midi-channel-ab.c
```diff
@@ -0,0 +1,217 @@
...
+
+    return &param;
+
+    // unused
+    (void)handle;
+}
+
+static float midichanab_get_parameter_value(NativePluginHandle handle, uint32_t index)
+{
+    if (index > MAX_MIDI_CHANNELS)
```
falkTX: Same, >=

source/native-plugins/midi-channel-ab.c
```diff
@@ -0,0 +1,217 @@
...
+        return 0.0f;
+
+    return handlePtr->channels[index] ? 1.0f : 0.0f;
+}
+
+static void midichanab_set_parameter_value(NativePluginHandle handle, uint32_t index, float value)
+{
+    if (index > MAX_MIDI_CHANNELS)
```
falkTX: Same, >=

source/native-plugins/midi-channel-ab.c
```diff
@@ -0,0 +1,217 @@
...
+        if (MIDI_IS_CHANNEL_MESSAGE(status))
+        {
+            const uint8_t channel = (uint8_t)MIDI_GET_CHANNEL_FROM_DATA(midiEvent->data);
+
+            if (channel >= MAX_MIDI_CHANNELS)
+                continue;
+
+            if (channels[channel])
+            {
+                tmpEvent.port    = midiEvent->port+1;
```
falkTX: this part is needlessly extensive.
please do a `memcpy(&tmpEvent, midiEvent, sizeof(NativeMidiEvent));` and then only modify the bits needed.

source/native-plugins/midi-channel-ab.c
```diff
@@ -92,18 +91,18 @@ static const NativeParameter* midichanab_get_parameter_info(NativePluginHandle h
...
 }
 
 static void midichanab_set_parameter_value(NativePluginHandle handle, uint32_t index, float value)
 {
-    if (index > MAX_MIDI_CHANNELS)
+    if (index >= MAX_MIDI_CHANNELS)
         return;
 
-    handlePtr->channels[index] = (value >= 0.5f);
+    handlePtr->channels[index] = (value >= 1);
```
falkTX: why did you change all floats to integers here?
the actual type is a float, so the *.0f makes sense, it is easy to understand when reading the code that the actual type is a float and not an integer.
The specific change that this comment replies to, please change it back to >= 0.5f.

milkmiruku: Ah, thank you. I'm not familiar with how to read C++.

source/native-plugins/_data.all.cpp
```diff
@@ -83,6 +83,22 @@ static const NativePluginDescriptor sNativePluginDescriptors[] = {
...
     DESCFUNCS
 },
+{
+    /* category  */ NATIVE_PLUGIN_CATEGORY_UTILITY,
+    /* hints     */ NATIVE_PLUGIN_IS_RTSAFE,
+    /* supports  */ NATIVE_PLUGIN_SUPPORTS_EVERYTHING,
+    /* audioIns  */ 0,
+    /* audioOuts */ 0,
+    /* midiIns   */ 1,
+    /* midiOuts  */ 1,
```
falkTX: Typo here, needs to be 2, not 1

source/native-plugins/_data.base.cpp
```diff
@@ -67,6 +67,22 @@ static const NativePluginDescriptor sNativePluginDescriptors[] = {
...
     DESCFUNCS
 },
+{
+    /* category  */ NATIVE_PLUGIN_CATEGORY_UTILITY,
+    /* hints     */ NATIVE_PLUGIN_IS_RTSAFE,
+    /* supports  */ NATIVE_PLUGIN_SUPPORTS_EVERYTHING,
+    /* audioIns  */ 0,
+    /* audioOuts */ 0,
+    /* midiIns   */ 1,
+    /* midiOuts  */ 1,
```
falkTX: Same, needs to be 2

source/native-plugins/_data.all.cpp
```diff
@@ -83,6 +83,22 @@ static const NativePluginDescriptor sNativePluginDescriptors[] = {
...
+    /* supports  */ NATIVE_PLUGIN_SUPPORTS_EVERYTHING,
+    /* audioIns  */ 0,
+    /* audioOuts */ 0,
+    /* midiIns   */ 1,
+    /* midiOuts  */ 2,
+    /* paramIns  */ 0,
+    /* paramOuts */ 0,
+    /* name      */ "MIDI Channel A/B",
+    /* label     */ "midichanab",
+    /* maker     */ "falkTX",
```
falkTX: You have maker here as me, but the actual plugin code has it as "Milk Brewster".
Which one is it?
If you want to set yourself as the author, that is completely fine, but please change in all places on the PR and Update the header comments too regarding it.
No need for "Copyright (C) 2012-2015 Filipe Coelho", just use your usual one-liner for these things.
I assume you're okay with GPL v2+ ?

milkmiruku: Thanks for catching that. Unless I'm mistaken, it's corrected below.

I've left your copyright header in midi-channel-ab.c also, given it's based on your code. Should I for sure be adding a line to files other than midi-channel-ab.c? I don't mind either way.

I'm fine with GPL v2+

falkTX:
Looks good, will merge now, thanks!

PR #642, opened 2018-05-31
Fix unsecured url in gidmodules

nick87720z:


nick87720z:
misprinting... gitmodules))

PR #644, opened 2018-06-01
Cut connections by Control+MButton3

nick87720z:
Simple click removes connections, which are too close to pointer
(for now 1px, but would be nice to make configurable).
Dragging cuts all connections on the way.

### DIFF
--- a/source/patchcanvas.py
+++ b/source/patchcanvas.py
@@ -24,6 +24,8 @@ from carla_config import *
 # ------------------------------------------------------------------------------------------------------------
 # Imports (Global)
 
+from math import floor
+
 if config_UseQt5:
     from PyQt5.QtCore import pyqtSignal, pyqtSlot, qCritical, qFatal, qWarning, Qt, QObject
     from PyQt5.QtCore import QAbstractAnimation, QLineF, QPointF, QRectF, QSizeF, QSettings, QTimer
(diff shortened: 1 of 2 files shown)


source/patchcanvas.py
```diff
@@ -1212,6 +1214,8 @@ def __init__(self, parent, view):
         self.m_ctrl_down = False
         self.m_mouse_down_init = False
         self.m_mouse_rubberband = False
+        self.m_mouse2_down = False
```
SpotlightKid: This attribute name is a bit ambiguous, don't you think? Are mouse buttons counted from zero or one?

source/widgets/draggablegraphicsview.py
```diff
@@ -40,14 +40,15 @@ def __init__(self, parent):
         QGraphicsView.__init__(self, parent)
 
         self.fPanning = False
+        self.m_ctrl_down = False
```
falkTX: Please rename this var to fCtrlDown, to be consistent with the rest of the code on this file.

source/patchcanvas.py
nick87720z: Just looked to docs about Qt::MouseButtons - indeed, that use special names for first several buttons, while numbered names refer to extra buttons, probably on gamer mice.
But still i don't remember cases, where numbering begins from 0.
For example, in gtk and some x11 apps (remembering configuring fluxbox) it begins with 1.
Anyway, i will rename it to match to qt docs.

nick87720z: Completely missed misleading MButton3 :/

PR #646, opened 2018-06-02
Lines hinting

nick87720z:
Here are changes, which fix look for antialiased borders. It is enabled for both canvas and minimap, slightly changing look for map image, while 2d cursor gets same antialiasing fix.

Note: this branch is based on mouse-connection-cut, which was confirmed to be ok, but is still opened.

### DIFF
--- a/source/patchcanvas.py
+++ b/source/patchcanvas.py
@@ -2782,6 +2782,8 @@ class CanvasBox(QGraphicsItem):
     def mouseReleaseEvent(self, event):
         if self.m_cursor_moving:
             self.setCursor(QCursor(Qt.ArrowCursor))
+            self.setX(round(self.x()))
+            self.setY(round(self.y()))
         self.m_mouse_down = False
         self.m_cursor_moving = False
         QGraphicsItem.mouseReleaseEvent(self, event)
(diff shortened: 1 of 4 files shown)


source/patchcanvas.py
```diff
@@ -1212,6 +1214,8 @@ def __init__(self, parent, view):
         self.m_ctrl_down = False
         self.m_mouse_down_init = False
         self.m_mouse_rubberband = False
+        self.m_mid_button_down = False
+        self.pointer_border = QRectF(0.0, 0.0, 1.0, 1.0)
```
falkTX: keep variable names consistent. rename this to "m_pointer_border".

source/patchcanvas_theme.py
```diff
@@ -337,14 +338,14 @@ def __init__(self, idx):
             self.port_font_state = QFont.Normal
             self.port_mode = self.THEME_PORT_SQUARE
 
-            self.port_audio_jack_pen = QPen(QColor(35, 61, 99), 0)
```
falkTX: why are you removing the custom colors?

falkTX:
I left 2 comments.
also, please rebase against latest master.

source/patchcanvas_theme.py
nick87720z: Please look above: this line is just moved out of theme-specific block. Color doesn't make sence, since it is placeholder for invisible line. Unfortunally, simple QPen(Qt.NoPen) has linewidth 1px, and alternative constructors require color, brush or like in begining, so i used "dumb" color.

nick87720z:
hm, it seems to be really problem with one branch, continuing another non-master. I will probably, have to reupload branch, may be recreate PR.

PR #647, opened 2018-06-03
Lines hinting

nick87720z:
Here are changes, which fix look for antialiased borders. It is enabled for both canvas and minimap, slightly changing look for map image, while 2d cursor gets same antialiasing fix.

Note: this branch is based on mouse-connection-cut, which was confirmed to be ok, but is still opened.


falkTX:
Thanks!!

PR #649, opened 2018-06-05
Code optimization

nick87720z:
Few small code optimizations and some refactoring in bounds of single functions.


nick87720z:
Oh, nice. As for middle corner i made it so, that regardless of corner angle, its line tries to go through pixel centers.

PR #652, opened 2018-06-09
Fix build with the clang compiler

yurivict:


falkTX:
Oh, so this is the only change needed?

yurivict:
Yes, the only one.

falkTX:
Alright, thanks for the report and fix.

PR #660, opened 2018-06-18
Do not use x86 asm with arm64

Johnnynator:
I don't know arm64 (or x86 asm) but the build does error with
```
../../modules/water/memory/ByteOrder.h:157:42: error: impossible constraint in 'asm'
     asm("bswap %%eax" : "=a"(n) : "a"(n));
```
which is probably a indication that this is not arm64 assembly :D

### DIFF
--- a/source/modules/water/memory/ByteOrder.h
+++ b/source/modules/water/memory/ByteOrder.h
@@ -153,7 +153,7 @@ inline uint32 ByteOrder::swap (uint32 n) noexcept
 {
    #ifdef CARLA_OS_MAC
     return OSSwapInt32 (n);
-   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || defined (__arm64__))
+   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || defined (__arm64__) || defined (__aarch64__))
     asm("bswap %%eax" : "=a"(n) : "a"(n));
     return n;
    #else
--- a/source/includes/CarlaDefines.h
+++ b/source/includes/CarlaDefines.h
@@ -60,7 +60,7 @@
 # define CARLA_OS_UNIX
 #endif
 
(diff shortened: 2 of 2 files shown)


source/modules/water/memory/ByteOrder.h
```diff
@@ -153,7 +153,7 @@ inline uint32 ByteOrder::swap (uint32 n) noexcept
 {
    #ifdef CARLA_OS_MAC
     return OSSwapInt32 (n);
-   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || defined (__arm64__))
+   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || ! defined (__arm64__))
```
falkTX: hmm this logic seems completely wrong...
how does this even work.

before we had, in pseudo-code:
if WINDOWS or not (arm or arm64)

now changed to:
if WINDOWS or not (arm or not arm64)

seems completely to go against the title you put on this PR.

Johnnynator: You are right. Now I'm confused why it errored at the next line when compiling for arm64 and now it does compile fine.

source/modules/water/memory/ByteOrder.h
```diff
@@ -153,7 +153,7 @@ inline uint32 ByteOrder::swap (uint32 n) noexcept
 {
    #ifdef CARLA_OS_MAC
     return OSSwapInt32 (n);
-   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || defined (__arm64__))
+   #elif defined(CARLA_OS_WIN) || ! (defined (__arm__) || defined (__aarch64__) || defined (__arm64__))
```
falkTX: nit-picking here, but please put the aarch64 check after the arm64 one.
so we check for arm, arm64 and then aarch64.
thanks

Johnnynator: OK. Done

falkTX:
Looks good, thanks!

PR #661, opened 2018-06-18
Only use GNU-specific char *strerror on systems with GLIBC

Johnnynator:
`char *strerror_r(int errnum, char *buf, size_t buflen);` is only mostly not available on linux systems without GLIBC. (e.g. Alpine linux or Void Linux with musl)


falkTX:
hmm this has to go into hylia or preferably ableton link repo directly.

See https://github.com/falkTX/Hylia and https://github.com/Ableton/link

falkTX:
The PR is appreciated though.

Issue #360, opened 2016-02-15
wish : confirmation dialog for "remove all"


Houston4444:
Today while working with carla, I wanted to add a plugin with the right click. I was a little too fast and I have clicked on "Remove All".
It wasn't too serious for me today, but it can be very frustrating if user didn't saved anything.
So, a simple confirmation dialog would be welcome.

falkTX:
I sorta agree.

I was thinking of showing a dialog if there's more than 1 plugin loaded.
Although it might be a little bit confusing why it doesn't show when there's 1 single plugin, so the dialog would have to mention it like "you have more than 1 plugin loaded, you sure you want to remove them all?".

How does that sound?

Houston4444:
I think that the best is to show dialog ever, this way there is no confuse. User who wants to remove the only one plugin quickly just has to right click on and select 'remove'.

Because 'remove all' is near from the cursor at the apparition of context menu, click accidentally on it is quickly done, so , I'd prefer a warning dialog in all cases.

unfa:
I have recent blew up some changes I made to my project with this feature.

I agree - let's warn every time, unless the canvas is empty :)

pdesaulniers:
This can be closed now.

PR #662, opened 2018-06-25
Add confirmation dialog for Remove All and New File

pdesaulniers:
Implements https://github.com/falkTX/Carla/issues/360. It prompts for confirmation even if there is only one plugin loaded. 

I hope I did that right :)

### DIFF
--- a/source/carla_host.py
+++ b/source/carla_host.py
@@ -1016,7 +1016,7 @@ class HostWindow(QMainWindow):
         if QMessageBox.question(self, self.tr("Remove All"),
                                       self.tr("Are you sure you want to remove all plugins?"),
                                       QMessageBox.Yes|QMessageBox.No) == QMessageBox.No:
-                                        return
+            return
 
         self.slot_pluginRemoveAll()
 
--- a/source/carla_host.py
+++ b/source/carla_host.py
@@ -416,7 +416,7 @@ class HostWindow(QMainWindow):
 
         self.ui.act_plugin_add.triggered.connect(self.slot_pluginAdd)
         self.ui.act_plugin_add2.triggered.connect(self.slot_pluginAdd)
-        self.ui.act_plugin_remove_all.triggered.connect(self.slot_pluginRemoveAll)
+        self.ui.act_plugin_remove_all.triggered.connect(self.slot_confirmRemoveAll)
 
         self.ui.act_add_jack.triggered.connect(self.slot_jackAppAdd)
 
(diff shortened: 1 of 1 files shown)


source/carla_host.py
```diff
@@ -1008,6 +1008,11 @@ def slot_pluginRemoveAll(self):
         if self.fPluginCount == 0:
             return
 
+        ask = QMessageBox.question(self, self.tr("Remove All"), self.tr("Are you sure you want to remove all plugins?"), QMessageBox.Yes|QMessageBox.No)
```
falkTX: multiple lines here please, and no need to save the return value, just compare directly to the No option.

falkTX:
open to thought, should the dialog be opened even if only 1 plugin is loaded?
perhaps only on >= 2?

also, is there any code in carla calling `slot_removeAll()` that would trigger this warning when not wanted?
(like changing projects in NSM)

pdesaulniers:
>  is there any code in carla calling slot_removeAll() that would trigger this warning when not wanted?

Ah, yes, it's called by the "New File" and "Open" actions. It's sort of appropriate for the New File action, but not for Open.

source/carla_host.py
pdesaulniers: Maybe it's just me, but I think this looks a bit less clear without saving the return value:
```
if QMessageBox.question(self, self.tr("Remove All"), 
                              self.tr("Are you sure you want to remove all plugins?"), 
                              QMessageBox.Yes|QMessageBox.No) == QMessageBox.No:
                                            return
```

pdesaulniers:
> should the dialog be opened even if only 1 plugin is loaded? perhaps only on >= 2?

This wouldn't be a huge time-saver :p 

By showing the dialog all the time, you get consistent behavior, and can possibly save a bit of frustration.

source/carla_host.py
falkTX: I still prefer this version tbh

pdesaulniers:
Seems OK now. ```slot_fileNew ``` seems to be only called by the menu action. Same thing for the new ```slot_confirmRemoveAll```.

source/carla_host.py
```diff
@@ -1003,6 +1011,18 @@ def slot_pluginAdd(self):
...
+
+        if QMessageBox.question(self, self.tr("Remove All"),
+                                      self.tr("Are you sure you want to remove all plugins?"),
+                                      QMessageBox.Yes|QMessageBox.No) == QMessageBox.No:
+                                        return
+
+        self.slot_pluginRemoveAll()
+
     @pyqtSlot()
     def slot_pluginRemoveAll(self):
```
falkTX: Since this function is no longer being used as a connection slot, better to remove the decoration and prefix from it.

falkTX:
Looks great!
Perhaps I would change the wording on those questions, but a really minor thing and can be done later anytime.
Thanks!

Issue #399, opened 2016-05-19
How do I Enable warn before closing?


RubenK77:
I was wondering how I can get a dialog box that asks me if I'm shure I want to close carla. Because it's really annoying if you are in a flow and accidently close carla.

pdesaulniers:
This can be closed now.

falkTX:
Yes, thank you.

PR #663, opened 2018-06-25
 Add confirmation dialog for quitting Carla; can be disabled in settings

pdesaulniers:
Implements https://github.com/falkTX/Carla/issues/399. I'm really not sure if I've done this correctly.


falkTX:
I appreciate the initiative, but there are some small things to note:

1. UI file changed in a few places without reason, did you use Qt5 designer?
For now we have to use Qt4 designer version as to make it compatible with Qt4 builds

2. using `self.fCustomStopAction` is a little unclear to me. I know it is my fault by having it there as raw ints to begin with, but we will have to change it to at least class-defined integer values so we know what the code is supposed to do.

3. saving `confirmExit` in the host object is unwanted, as it has nothing to do with the backend.
the cached value needs to go into `self.fSavedSettings` under `loadSettings()` instead

4. did you verify that using ctrl+c does NOT trigger the question?
did you verify changing projects using NSM? I dont think it triggers a close event, but just to be sure.

resources/ui/carla_settings.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>612</width>
-    <height>544</height>
+    <width>600</width>
```
falkTX: Please change this back to original size.
Can be done manually.

resources/ui/carla_settings.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>612</width>
-    <height>507</height>
+    <width>600</width>
+    <height>587</height>
```
falkTX: Please change this back to original size.
Can be done manually.

source/carla_host.py
```diff
@@ -2281,6 +2282,13 @@ def timerEvent(self, event):
     # close event
 
     def closeEvent(self, event):
+        if not self.fCustomStopAction == 1 and self.fSavedSettings[CARLA_KEY_MAIN_CONFIRM_EXIT]:
```
falkTX: This should be "action != 1" rather than "not action == 1" :)

pdesaulniers:
Regarding fCustomStopAction, there's a comment that says: 
```        
# run a custom action after engine is properly closed
# 1 for close carla, 2 for load project
```
I've used it so that ctrl+c does not trigger the question. Seems like it should be ```action == 0``` instead. Or is that variable inappropriate to use here?

falkTX:
Using != 1 seems correct. It is needed to ignore ctrl+c events.
There should be a check for == 2, to give a different dialog if closing while a project is still loading, but that can be for another time.

source/carla_host.py
```diff
@@ -2281,6 +2282,13 @@ def timerEvent(self, event):
...
+            ask = QMessageBox.question(self, self.tr("Quit"), self.tr("Are you sure you want to quit Carla?"), QMessageBox.Yes|QMessageBox.No)
```
falkTX: small nitpick. Line is too long here.
often I align them with the start of the first `self.tr()`, so the 2nd one looks similar.
not all code in carla is good space-wise, but I dont want to add more to it..

pdesaulniers:
> There should be a check for == 2, to give a different dialog if closing while a project is still loading

Why would this be useful?

> did you verify changing projects using NSM?

To be honest, I'm not super familiar with NSM, so I can't tell for sure whether or not everything is running fine right now. The prompt doesn't show up when switching sessions, at least.

> using self.fCustomStopAction is a little unclear to me [...] we will have to change it to at least class-defined integer values

Such as CUSTOM_STOP_ACTION_CLOSE_CARLA and CUSTOM_STOP_ACTION_NSM_OPEN? (please bear with me :p)

falkTX:
Some enums like that yeah, but code is fine already.
A bit dangerous to merge now since I want to do a bugfix release, but it is still a beta one, so why not just go forward.

pdesaulniers:
Maybe it would have been good to disable it by default?

falkTX:
I dont mind it being on by default.
Might be expected user experience.

After RC is out, these changes wont be allowed anymore.
But for now it is fine.

PR #700, opened 2018-08-26
Set font size for plugin preset and midi/audio indicator labels

net147:
Fixes text for plugin knobs being clipped in some desktop environments.


falkTX:
I did not have the issue myself, the changes do "nothing" for me, so all good :)
Thanks.

PR #709, opened 2018-08-31
Add CarlaVstShell

jhernberg:
Add VST Shell capability.

### DIFF
--- a/source/plugin/Makefile
+++ b/source/plugin/Makefile
@@ -170,7 +170,7 @@ $(BINDIR)/CarlaPatchbayFX$(LIB_EXT): $(OBJDIR)/carla-vst.cpp.patchbay-fx.o $(LIB
 	@echo "Linking CarlaPatchbayFX$(LIB_EXT)"
 	@$(CXX) $< $(LIBS_START) $(LIBS) $(LIBS_END) $(SHARED) $(LINK_FLAGS) -o $@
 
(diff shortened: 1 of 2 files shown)


source/plugin/Makefile
```diff
@@ -169,6 +170,11 @@ $(BINDIR)/CarlaPatchbayFX$(LIB_EXT): $(OBJDIR)/carla-vst.cpp.patchbay-fx.o $(LIB
 	@echo "Linking CarlaPatchbayFX$(LIB_EXT)"
 	@$(CXX) $< $(LIBS_START) $(LIBS) $(LIBS_END) $(SHARED) $(LINK_FLAGS) -o $@
 
+$(BINDIR)/CarlaVstShell$(LIB_EXT): $(OBJDIR)/carla-vst.cpp.vst-shell.o $(LIBS)
```
falkTX: small nitckpick, I would name this object carla-vst.cpp.shell.o, taking out the -vst suffix.
It is already implied it is a vst :)

jhernberg: OK, I'll do that.

falkTX:
I am going to close this since CarlaVstShell is now always built by default.
It does not get installed, as it is not useful for now. But it works and can load carla as plugin in shell mode.
I made it work as a wine plugin too, so we can run linux-native carla inside a windows daw running in wine :)
(run `make plugin-wine`)

PR #751, opened 2018-09-18
Add X _NET_WM_WINDOW_TYPE flags to plugin windows 

diovudau:
Add X _NET_WM_WINDOW_TYPE flags to plugin windows so they work equally well on tiling window manages and floating WM.

Tested with QTile 2869.95919912 and XFCE 4.12.1

I consider this a trivial change which will not create any copyright for me.


falkTX:
Thanks :)
The same should be done for DPF, but I can just copy&paste the code there.
Will create a ticket for it

PR #754, opened 2018-09-23
Tracking of 2.1 changes

falkTX:
PR so we can more clearly see what changes have been added into develop that did not get cherry-picked into master.
This will facilitate making the changelog for v2.1


falkTX:
Changes so far:
- reintroduction of juce, used for windows and mac builds. this brings AU and VST3, plus better audio device support. linux and all other open OSes never touch juce, leaving them supported platforms
- removal of PyQt4 for frontend, now requires PyQt5
- patchcanvas code rework, lots of small handy changes
- VST plugin now uses dummy window to start carla, X11 embed tricks removed
- VST plugin now also available in macOS and Windows builds
- Wine VST version, to load Linux plugins inside Windows applications
- Resizable LV2 X11 UIs
WIP: LV2 param support

source/backend/plugin/CarlaPluginNative.cpp
```diff
@@ -2481,7 +2479,7 @@ class CarlaPluginNative : public CarlaPlugin
             return false;
         }
 
-        if (label == nullptr && label[0] != '\0')
+        if (label == nullptr || label[0] != '\0')
```
AndiEcker: label[0] == '\0'

falkTX: ah yes!

falkTX:
rough 2.1 changes:

drop qt4 support
canvas connect action
small canvas, AA, cut connections, rework nick87720z 
colorize all slot skins?
internal jack client
Dynamically show knobs in rack view as much as possible
carla-control and osc rework
juce stuff again, but disabled by default on source builds, etc
Implement clearing xrun count, bottom bar stuff
Save&restore Session Manager type in add-jack-app dialog
NSM support for JACK Applications (#829)
Introducing "Cancelable actions"
Initial support for LV2 params (#811)
Ship vst plugin in windows builds
Update READMEs to mention VST3 and AU support
Working wine-native bridge, needs testing
Respect X11 UIs size hints from plugins
Build CarlaVstFxShell by default?
rework osc, carla-control full
libjack stuff
internal client
testing carla-vst exposed parameters
inline display in canvas

PR #766, opened 2018-10-11
Fixes for fluidsynth >= 2.0.0 V2

schnitzeltony:
It needs a new pull request for develop branch. Will close the other one

V1 -> V2:
* rebase against develop branch
* rename function & vars
* add missing 'sbFluidDefaultsStored = true;'

### DIFF
--- a/source/backend/plugin/CarlaPluginFluidSynth.cpp
+++ b/source/backend/plugin/CarlaPluginFluidSynth.cpp
@@ -73,20 +73,22 @@ public:
         fSynth = new_fluid_synth(fSettings);
         CARLA_SAFE_ASSERT_RETURN(fSynth != nullptr,);
 
+        initializeFluidDefaultsIfNeeded();
+
         fluid_synth_set_sample_rate(fSynth, (float)pData->engine->getSampleRate());
 
         // set default values
         fluid_synth_set_reverb_on(fSynth, 1);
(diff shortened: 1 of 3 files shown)


source/backend/plugin/CarlaPluginFluidSynth.cpp
```diff
@@ -1683,13 +1770,19 @@ class CarlaPluginFluidSynth : public CarlaPlugin
     float** fAudio16Buffers;
     float   fParamBuffers[FluidSynthParametersMax];
 
+    static bool  sbFluidDefaultsStored;
```
falkTX: I meant before to use "s" prefix *instead* of "b" or "f", just a single one.
Rest looks quite fine, thanks!

source/backend/plugin/CarlaPluginFluidSynth.cpp
```diff
@@ -1656,6 +1689,60 @@ class CarlaPluginFluidSynth : public CarlaPlugin
     }
 
 private:
+    void initializeFluidDefaultsIfNeeded()
+    {
+        if (!sbFluidDefaultsStored)
```
falkTX: please revert this check, like so:
```
if (sFluidDefaultsStored)
    return;
sFluidDefaultsStored = true;
// ...
```
The purpose is to continue the main code without extra indentation

schnitzeltony:
V3: 
* rebased to develop branch
* variable rename
* if bails out in case default were loaded already

falkTX:
Very nice, thanks!

PR #768, opened 2018-10-12
Minimap enhancements

nick87720z:
Sidebar: 1px browser elements spacing
Minimap:
- better frame, created from tab-less tabwidget,
- update on each selection change,
- some color scheme awareness,
- full resize (vertical too),
- better cursor blending,
- mouse restriction during movement,
- scale mode.


falkTX:
Thanks again! :)

PR #770, opened 2018-10-12
Rack enhancements 2

nick87720z:
Same stuff as in deleted "rack-enhancements", but changed highlight side effect to side lines instead of pre-generated light&bloom images.

Can't even create new PR after existing is closed due to branch removal.


falkTX:
Merged manually via CLI, as github still somehow screwed up for me.
Thanks very much!

PR #774, opened 2018-10-14
Frontend fixes

nick87720z:


nick87720z:
Probably it was redundant - i made this change at the same time with selection fix, after got several rare segfaults with glow fx, attached to selected port.

PR #776, opened 2018-10-15
More reliable connection line (de)selection

nick87720z:
I still got sitation, when signal from deselected port comes after
selected one (i.e. select both ends with Ctrl, then select input end).


falkTX:
Fair enough, thanks

PR #777, opened 2018-10-15
Canvas tiny scale visibility

nick87720z:
Added additional cosmetic-type layer to connection lines and OpenOctave theme port lights, making these elements distinguishable on tiniest scales on minimap.
OpenOctave theme example


falkTX:
Not sure if I understand everything, but alright. Not going to stop you :)

PR #779, opened 2018-10-15
Fix/pgm as bank

bencrossman:


falkTX:
Please target the PR against develop, master is "closed" for the upcoming stable 2.0 release.
All commits should go to develop first, then cherry-picked into master.
Thanks

PS: Did you check if CarlaEngineJack.cpp needs the same fix?

bencrossman:
Done

No it seems this is the only occurrence of this bug. In fact I think CarlaEngineEventPort::writeMidiEvent could be simplified to just write out kEngineEventTypeMidi's but I don't have enough understanding of the code yet.

falkTX:
Alright, thanks

PR #489, opened 2017-06-14
changes to implement #244 - send bank and program change for vst2

hottwaj:
I've had a go at implementing changes to make bank and program change work for vst2 along the lines suggested in #244 and following similar code mentioned for LV2

As with others in #244, I have a win32 vst (AAS' Lounge Lizard) that provides several different banks (as well as programs in each bank) that can be changed via MIDI - currently not possible in Carla.

I can compile my changes, but I'm running 16.04 and can't build carla with win32 vst support as I can't install mingw32 (it's not available for xenial?), so my changes are untested (sorry).  BTW I'm following [this guide for compiling from source](https://www.linuxmusicians.com/viewtopic.php?t=12812) - hopefully that's appropriate!

I realise this is probably not a priority but I'm keen to help in any way I can to get this merged (or if you'd prefer not to merge, any help on compiling for xenial, if possible, would be appreciated!)

My needs are quite simple (I only need this plugin - for now!) so I'm probable using the wrong tools for the job - but I also can't get wineasio to work, whereas Carla works perfectly!  

Thanks for your work on this great set of tools :)

PR #782, opened 2018-10-16
Revised attempt at: https://github.com/falkTX/Carla/pull/489

bencrossman:


bencrossman:
Now that Carla 2.0 is out, can we merge this pull request?

Thanks

falkTX:
yes, but now we really need to put this back in the juce class since it is back once again :)
sorry

bencrossman:
ah great ok, will do that

bencrossman:
Ok, think this is good now

falkTX:
Sorry I got distracted and forgot about this.
It has been too long. Let's merge now and if there is an issue, we fix later.

bencrossman:
Thank you, awesome

jdarnell2:
I'm using a plugin, (Amplitube 5) that switches presets via Program Change messages.  How can I implement the above added feature into the latest version of Carla?  Do I have to rebuild and re-compile?  I'm using Windows 10 x64.

PR #810, opened 2019-01-11
Fixed shell export syntax

qwertysam:
Previously, running Carla from this script would give the error, likely because the shell interpreter is not understand the more advanced syntax. Now uses simpler syntax.

`/usr/bin/carla: 14: export: Desktop:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin: bad variable name`

Note: Alternatively, to use the previous syntax, the script must be using #!/bin/bash
(Tested on Ubuntu 18.04.1)

### DIFF
--- a/data/build-haiku.sh
+++ b/data/build-haiku.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 export HAIKU=true
 export CC=gcc-x86
--- a/data/carla
+++ b/data/carla
@@ -11,5 +11,6 @@ if [ "$1" = "--gdb" ]; then
 fi
 
 INSTALL_PREFIX="X-PREFIX-X"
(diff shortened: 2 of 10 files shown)


falkTX:
I never noticed this was a problem, thanks for the bug report.
Would you mind changing the other startup scripts to have the same?
There is carla-rack, carla-patchbay and a few others.

Note: Usually PRs should target the develop branch, but since this is a general problem it is okay to go into master first since I would end up cherry-picking the commit there anyway

qwertysam:
Looking at the other scripts, using Bash seems to be a better approach than rewriting all the syntax, so I've changed that.

falkTX:
Nice, thanks!

PR #815, opened 2019-01-20
Resurrect Juce

falkTX:
PR to remain open while I think if it is worth it...


falkTX:
This will now be merged, but juce is disabled by default.
It is only going to be enabled (later on) for Windows and macOS, as dealing with VST3 and AU is a big pain...

trebmuh:
So the idea is to make Carla able to deal with VST3 under Windows and macOS (but not under linux), is that right?

falkTX:
sorta. we leverage juce on those platforms, but leave linux fully free of bad libraries.
this is because I do not want to deal too much with closed/proprietary OSes too much, better to let someone else handle all the quirks that these damn OSes require...
I am not opposed to deal with VST3 on Linux, but juce does not support it yet and we have almost 0 vst3 linux plugins right now.
I would actually prefer if vst3 did not became a thing on linux. we have lv2 which is far superior...

trebmuh:
OK, fair enough.
Thanks (with a bit of latency) for the explanation Filipe

Issue #596, opened 2018-03-19
allowing internationalisation


trebmuh:
When it'll be possible, I'm willing to provide a French translation of it. So far, I'm patching against the sources.

_**Edit**: in case it wasn't clear enough, I've got a full translation in French already ready._

falkTX:
To be honest, I am kinda against this idea for Carla.
I appreciate your offer, but I rather not focus on this at all.

As Carla grows more complex and me having less time for it, eventually it will have to stay in a more maintenance mode.
I'd like to have it not so different from what it is now, just a lot more stable. (and supporting new systems as they come along, that should keep me busy on it).

So sorry, but I am closing this request.
Feel free to ask again in 2 years :)

trebmuh:
I'll try to keep your last sentence in my brain for the next 2 years then. We'll see :)

trebmuh:
Happy birthday #596. Only one year before I'm allowed to ask again! :wink:

falkTX:
haha, great to see the continued interest.
I am more relaxed on this now, but carla is not yet at a point where I am done adding stuff.
It is really almost there though!
Next year we can begin translation, I guess for v2.2

trebmuh:
:soon: \o/

trebmuh:
roooo, damn it! :wink:

allejok96:
Greetings from the LMMS project! It seems we have accidentally uploaded strings from the entire Carla codebase onto our Transifex page, and poor translators have translated it in vain. I've just begun looking into how to fix this. I don't know how many languages that have translations for the Carla strings... maybe 5-10. I just thought I'd let you know there are probably some resources that can be repurposed with a bit of clever regex magic.

PR #860, opened 2019-03-21
Add the frontend translation mechanism

jpcima:
#596
This adds the mechanism to build, install, load Qt translations into the frontend.

The desired language code is added to `I18N_LANGUAGES`. (example `fr_FR`)
This is the behavior to expect:
1. an initial `make` commands scans sources, creates `.ts` files (`pylupdate5` step)
2. the default build rule will generate `.qm` files
3. installing will copy `.qm` into `resources/translations`

Please note: no translation is included, nor is there a plan to add one shortly after this.
This is to help the job of translators to produce and maintain a custom translation of their making, like LibraZiK does.


falkTX:
Niice, you saved me a lot of work!

I won't merge this just now, as it is not a feature I want to support in v2.1
But after v2.0.0 is announced and branches get switched around (master will be incoming v2.1, and develop v2.2/next), then I will merge this in there.

There are a lot of special things in the makefile though.
Do you know if these will work on non-gnu systems? (like FreeBSD and MacOS)

jpcima:
>There are a lot of special things in the makefile though.
Do you know if these will work on non-gnu systems? (like FreeBSD and MacOS)

This would have to be checked, as I don't have familiarity with either of these Make systems.
I have tried only GNU/Linux.

jpcima:
I have had success on FreeBSD12 and macOS 10.12.

I'm revising the code in just a minor way: for the Qt part of translations, aka `qt_*.qm` files, I will perform in order the 2 loading attempts: 1. the case these files are shipped as carla resource, and 2. the case they are present in the Qt distribution.
This will allow macOS i18n support in two cases, the one of an app bundle, or a `make install` build from the sources.

About FreeBSD, I built with the tool GNU make `gmake`.
BSD make did not work in any case, whether with or without the translation work.

jpcima:
Rebased on `develop`, conflicts resolved

trebmuh:
Just mentioning. Note that, I've got one translation ready for French. No hurries but so you know.

falkTX:
Changes have been merged to develop branch. Scheduled for whenever 2.3 releases

falkTX:
and of course, thanks!

PR #864, opened 2019-03-21
Fix a build error on FreeBSD

jpcima:
Include <signal.h> in order to make `kill`, `SIGKILL` available.


falkTX:
Makes sense, thanks!

PR #887, opened 2019-06-08
Change the dialog icon used for cancelable actions

jpcima:
I find this icon :information_source: more fitting for the cancelable action, rather than using :x: which suggests error and it's a little confusing.


falkTX:
hmm yeah. even the warning one would be more appropriate, but info is fine too.
will merge now, thanks

PR #895, opened 2019-07-02
Fix Qt .moc include location for Qt >= 5.13

### LINKED ISSUE #894: Carla doesn't compile with gcc 9.1 and Qt 5.13
Trying to compile carla-git on ArchLinux gives the following:
[carla_gcc9_make-error.txt](https://github.com/falkTX/Carla/files/3349935/carla_gcc9_make-error.txt)

dvzrv:
source/bridges-ui/CarlaBridgeToolkitQt.cpp: Moving the include of CarlaBridgeToolkitQt{4,5}.moc to not be within CARLA_BRIDGE_UI namespace.

Fixes #894

### DIFF
--- a/source/bridges-ui/CarlaBridgeToolkitQt.cpp
+++ b/source/bridges-ui/CarlaBridgeToolkitQt.cpp
@@ -258,6 +258,21 @@ private:
 
 // -------------------------------------------------------------------------
 
+#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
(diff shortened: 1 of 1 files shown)


dvzrv:
Hm, odd, this works for me locally, with gcc 9.1.0 and qt 5.13

dvzrv:
Oh, I understand... it's actually the qt4 stuff that is now failing! I can fix that I guess.
But oh my... qt4 (maybe time to die? ;-) ).

terencode:
Confirmed working on latest commit from `develop` branch :+1:

PR #916, opened 2019-09-25
Fix source comment and user facing typos

luzpaz:
Found via `codepell` (v1.17.0.dev0)  
```
codespell -q 3 -L iff,iif,inport,sord,sinc,te,uint -S ./source/includes,./source/modules,./source/native-plugins,./data/windows
```


falkTX:
this is neat, thanks!

PR #921, opened 2019-10-01
fix missing - from rpath in pc files

alex-tee:
from IRC
```
<alextee> falktx, i get "/usr/bin/ld: cannot find rpath=/usr/lib/carla: No such file or directory" when using carla-native-plugin from pkgconfig
...
<alextee> falktx, i asked in #mesonbuild and it looks like the --libs of carla-native-plugin is wrong
<alextee> it currently says: -Wl,rpath=/usr/lib/carla -L/usr/lib/carla -lcarla_native-plugin
<alextee> but i was told it should be "-Wl,-rpath,/usr/lib/carla"
```

using `-Wl,-rpath,...` instead of `-Wl,rpath=...` made by build succeed

I think the problem is not the `=` but the missing `-` before rpath:
https://stackoverflow.com/a/15667612

EDIT:
changed to `-Wl,-rpath=dir`

### DIFF
--- a/data/carla-native-plugin.pc
+++ b/data/carla-native-plugin.pc
@@ -5,5 +5,5 @@ includedir=X-INCLUDEDIR-X/carla
 Name: carla-native-plugin
 Version: 1.9.14
 Description: Carla Native Plugin
-Libs: -Wl,rpath=${libdir} -L${libdir} -lcarla_native-plugin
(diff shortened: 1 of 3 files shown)


falkTX:
I just recently got hit by the same issue. very annoying errr, will merge now, thanks

PR #927, opened 2019-10-10
Canvas pairs, to test and talk

Houston4444:
That's here.
So, I didn't remember it represents so much code.
Sorry the code is quite dirty now, but that gives you an idea of how it works.

I named this "pair" to not be confused with groups, but it's not conceptually limited to stereo (even if there is no user way to make more than 2 ports pair).

Here stereo detection is set on (detected with port names), it can be set of commenting all the condition( if 0 == 0: line 1312).
There is also the possibility to set a port as stereo with a neighbor via contextMenu on port.
Possibility to split the pair via context menu on pair widget.

Pairs are saved in settings. Here there is something to decide,
I think it should be: save in settings + save in project (with priority to project, and no save in settings in carla patchbay).

Little thing more : Linemov becomes dotted hover a connected port (means it would disconnect).


terencode:
You need to rebase that against develop branch I think. It currently contains commits that are not related to this PR.

Houston4444:
terencode, I made a new PR #928 based on master branch.
This is yet more for test and talk about how to implement it neatly.
I'll remake a PR on 'develop' branch if needed when needed.

terencode:
As @falkTX pointed out, you need to target the develop branch for new features.

PR #928, opened 2019-10-10
Canvas pairs

Houston4444:
That's here.
Sorry I just made a PR to your 'develop' branch. I'll remove it.

So, I didn't remember it represents so much code.
Sorry the code is quite dirty now, but that gives you an idea of how it works.

I named this "pair" to not be confused with groups, but it's not conceptually limited to stereo (even if there is no user way to make more than 2 ports pair).

Here stereo detection is set on (detected with port names), it can be set of commenting all the condition( if 0 == 0: line 1312).
There is also the possibility to set a port as stereo with a neighbor via contextMenu on port.
Possibility to split the pair via context menu on pair widget.

Pairs are saved in settings. Here there is something to decide,
I think it should be: save in settings + save in project (with priority to project, and no save in settings in carla patchbay).

Little thing more : Linemov becomes dotted hover a connected port (means it would disconnect).

### DIFF
--- a/source/frontend/patchcanvas.py
+++ b/source/frontend/patchcanvas.py
@@ -66,6 +66,10 @@ PORT_TYPE_MIDI_JACK  = 2
 PORT_TYPE_MIDI_ALSA  = 3
 PORT_TYPE_PARAMETER  = 4
 
+#Port Stereo position
+PORT_IN_PAIR_POSITION_MONO  = None
+PORT_IN_PAIR_POSITION_LEFT  = 0
+
 # Callback Action
 ACTION_GROUP_INFO       =  0 # group_id, N, N
 ACTION_GROUP_RENAME     =  1 # group_id, N, N
@@ -217,6 +221,7 @@ class Canvas(object):
         'last_connection_id',
         'initial_pos',
         'size_rect',
+        'is_line_mov',
         'group_list',
         'port_list',
         'force_mono_list',
@@ -1384,7 +1389,6 @@ def CanvasGetNewPairPortIdList(group_id, port_id, port_name, port_mode, port_typ
                             canvas.port_list.insert(index+1, last_port)
                             return [ port.port_id, port_id ]
                             
-        
 def CanvasAddPair(group_id, port_mode, port_type, port_id_list):
     box_widget  = None
     pair_widget = None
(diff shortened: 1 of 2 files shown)


falkTX:
Thanks a lot, but please dont target master branch as that one is for stable/fixes only.
develop is the default branch for good reasons.

Before I merge any of this, note that grouping is to be done on the backend side, providing a "key" for each group. This will fit nicely into jack and lv2 specs

Houston4444:
I will remake my PR against develop branch but as you said "grouping is to be done on the backend side".
Do you plan to do that, I don't feel really able to. (I failed to find a jack port specification for that in jack documentation and source code).

falkTX:
backend side I can take care of, of course.
it is easier to do first with lv2 plugins, since we already have a few that use that.
jack meta-data is too new

Houston4444:
Ok, thanks.
So there is something I don't understand.
Say me if I am wrong, but I guess that in backend (except for carla-patchbay), this is when a JACK port appears that a port appears in canvas patchbay. I guess in the case of a LV2 plugin, Carla creates JACK ports for each lv2:InputPort or lv2:OutputPort and then these JACK ports are added to patchbay. So if we don't take care of JACK metadata, how can this information be sent to GUI ?

Something else: where should be placed a stereo detection with port names (obviously only active if we've got no information about Port Groups (that I called Pairs) ?

falkTX:
That part still has to be defined. I am in the middle of a big commit so won't push right away, but one easy way I see we can do this is by using the `valueStr` on `ENGINE_CALLBACK_PATCHBAY_CLIENT_DATA_CHANGED` to mean group name.
With a note that groups are per client/plugin.

falkTX:
ah no, the `ENGINE_CALLBACK_PATCHBAY_CLIENT_DATA_CHANGED` is for clients.
we need something that applies to a port. let me see..

Houston4444:
humm, if client means group, I would say it can be fine.
The stereo detection could be applied only on clients which didn't send any port-group information.
Patchcanvas could make pairs once all ports for the pair are here, or right away if ports exists.
The problem I see is that ENGINE_CALLBACK_PATCHBAY_CLIENT_DATA_CHANGED has to send all the port-groups of the client.

falkTX:
API for port groups added in 0f497f5f9a9fc90d04cff630fe18041e1f37a4d5
No implementation for this yet, but that should be able to get us moving a bit for now..

Houston4444:
Good.
I think to something, I wonder if it is really a good idea to separate input/output of the port-group before patchcanvas.py . There are maybe nice things to do in canvas if it could know which input matchs with which output.

falkTX:
I followed lv2 spec regarding those hints. I am okay with changes there, I just wanted to have some minimal hints

btw, the commit already has in place empty stub functions where you can plug your code to handle groups.
let me know if you need something extra

Houston4444:
Yes Ok thanks.
I can work with that. I think my PR for canvas could be ready in 2/3 days.

Houston4444:
see #935

PR #929, opened 2019-10-11
Add more native host opcodes to callback.

CYBERDEViLNL:
I hope the opcodes are mapped as intended? And I let it return if the `pluginId != 0` since we only handle the first plugin, is this right? 

Also I don't know how to map `NATIVE_HOST_OPCODE_UI_TOUCH_PARAMETER` and `NATIVE_HOST_OPCODE_INTERNAL_PLUGIN`.

### DIFF
--- a/source/backend/engine/CarlaEngineNative.cpp
+++ b/source/backend/engine/CarlaEngineNative.cpp
@@ -322,10 +322,61 @@ public:
     {
         CarlaEngine::callback(sendHost, sendOsc, action, pluginId, value1, value2, value3, valuef, valueStr);
 
-        if (action == ENGINE_CALLBACK_IDLE && ! pData->aboutToClose) {
-            pHost->dispatcher(pHost->handle,
+        // TODO?
+        // We currently only have params from the first plugin.
+        if (pluginId != 0) { return; }
+
+        switch (action) {
+            // TODO
+            // NATIVE_HOST_OPCODE_UI_TOUCH_PARAMETER
+            // NATIVE_HOST_OPCODE_INTERNAL_PLUGIN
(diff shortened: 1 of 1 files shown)


falkTX:
You are mapping way more than what is needed, without knowing why, please don't.
A few things:
1. carla maps all plugin parameters until it reaches the 100 number, it is not just the first plugin
2. MIDI cc/channel and program change does not apply here, the native-api message is not related to the engine one
3. `void callback` is there my mistake, this code should go into `uiServerCallback` instead (though I am not 100% sure of this)
4. param changes are already handled without catching the callback, no?
5. inline display and idle are a good catch, I forgot about these. the ui touch one is done via dispatcher call (it is not its own function in order to keep ABI compatibility with old versions of carla)

falkTX:
note I am confused here why the engine callback override is there, and what/when uiServerCallback is being called then.

uiServerCallback is likely to send host related messages to the remote UI.
so maybe it makes sense to have this separated.

note that NATIVE_HOST_OPCODE_UPDATE_* is to tell the host to update the metadata/labels of something. like updating the gui rendering of a specific parameter, that changed for example from "unused" to "gain".

CYBERDEViLNL:
Ok never mind this change, it all started working without this change. (I wasn't using develop branch before.) :unamused:

It only works for the first plugin. Maybe this has something todo with that `CarlaPlugin* _getFirstPlugin()` is called a few times in the code? https://github.com/falkTX/Carla/blob/develop/source/backend/engine/CarlaEngineNative.cpp#L1722

falkTX:
That seems indeed to be the case, I must be confused with something else, I thought I implemented support for more than just the first plugin...

PR #935, opened 2019-10-18
Canvas Port groups

Houston4444:
Ok, this is it !
concerning the API, 

I changed portId to portFlags in PortGroupAdded arguments, because I need to know the portMode and portType (maybe it could be used for non-audio, why not).
I don't use PortGroupChanged because the PortGroup name is set by its ports names.
I did change 'renamePort' to 'changePortProperties' in patchcanvas.py

Port Group has to be added to patchcanvas before adding the concerned ports.
port group widget is created once it contains two ports, so please check if is widget is not None before to call it.
Port group removes from him all it port_ids when one a its ports has been removed.
Remove a Port Group do not removes its ports (but set them out of a port group). 

I added two canvas callbacks, one for Adding a Port Group (from Canvas, right click on a port), one for Remove a Port Group (right click on port group).
These are for now treated directly in canvas_host.py , because it was simpler for the moment. Note that PortGroups are not saved, this is something to decide (Save globally in config files, save in project file, both...). 

There is no Stereo Detection in this PR !

Good New: I've got no more segmentation faults at split/join groups.
Bad new: I don't know exactly why ;).

There are certainly many many things to say, check me on IRC if needed.


source/frontend/carla_host.py
```diff
@@ -1618,18 +1618,34 @@ def slot_handlePatchbayPortRemovedCallback(self, groupId, portId):
...
     def slot_handlePatchbayPortChangedCallback(self, groupId, portId, portFlags, portGroupId, newPortName):
-        patchcanvas.renamePort(groupId, portId, newPortName)
+        patchcanvas.changePortProperties(groupId, portId, portGroupId, newPortName)
         self.updateMiniCanvasLater()
 
     @pyqtSlot(int, int, int, str)
-    def slot_handlePatchbayPortGroupAddedCallback(self, groupId, portId, portGroupId, newPortName):
-        # TODO
-        pass
+    def slot_handlePatchbayPortGroupAddedCallback(self, groupId, portFlags, portGroupId, newPortName):
```
falkTX: I dont understand why the change of argument order was needed here..

falkTX: oh wait, I understand, this is adding flags instead of port-id which seems useless for you.
please rename portFlags to portGroupFlags then, thanks

source/frontend/patchcanvas/__init__.py
```diff
@@ -142,6 +146,8 @@ def __init__(self):
         self.last_connection_id = 0
         self.initial_pos = QPointF(0, 0)
         self.size_rect = QRectF()
+        
+        self.is_line_mov = False
```
falkTX: is this variable actually needed?

falkTX:
left a few comments, it is quite big, will review later.
thanks a lot.

was the canvasbox2.py file added by mistake?

Houston4444:
canvasbox2.py was added by mistake, yes.

canvas.is_line_mov variable is not needed yet, you are right.
This variable is needed for another feature that I did implement, and I don't know if you want it.
This feature is quite simple: when user has pressed on a port (or port group), and he is moving the cursor to another port, he can directly makes the connection with a right click (keeping the left button pressed). This way, the connection line mov is still here and he can still connect the first port to other ports.

Say me if you want me to remove all that concerns this feature (the variable canvas.is_line_mov and other commented code).

falkTX:
I dont want to take forever on merging this. Life is too short to be always waiting for stuff.
I had the impression/idea that this would be tighed to the hints provided by the backend, but I see you made it also possible to manually change these groups. is that correct? (I did not try the PR yet)

The connection of ports is already possible in a similar way as you describe no?
In ingen someone can click on a output port, then input port and it just makes the connection.
Carla could just do the same. BUT it has to be opt-in.
I am thinking of adding a few small buttons on the top-right of the canvas, for behaviour-changing stuff. We already have such a thing for "hover selects ports" in the settings, which can go there instead.
The logs tab need a similar thing too, one button for "follow mode" and other to clear the log. But that is something else now...

In any case, one thing at a time.
Please remove the canvasbox2.py and `is_line_mov` variable, I will merge this afterwards.
Thanks

PS: do you mind if I squash the entire PR in a single commit?

Houston4444:
No problem. Before your message, reading the file changes, I was thinking it was too big and too much information at one time. 
The reason for this is that this parts of code are 3 years old (at this time, not working with git or anything else), and it was simpler for me now to adapt all the features from Catia(3 years ago) to Carla. So I am sorry for this, I will remove from this PR all what doesn't concerns directly the port groups.

This feature has not a big interest if we limit its use to the hints provided by the lv2 plugins which talk with host about port groups. In fact, if we have a port group for a lv2 plugin, but not with the system playback ports, we have no faster way to connect the stereo pair.
The backend could auto detect port groups with port names (In case there is no port group info from jack metadata or lv2), but it will works only in 99 % of cases. For the 1 % of cases, it's needed to provide an user way to group/ungroup the ports.
I can also remove "set as stereo/split to monos" features from the PR, but you will have no way to test port groups directly.

PS: no problem with one commit for all.

falkTX:
ok I understand, I will wait for the next PR then.

regarding manual groups, I see your point.
we will figure out a way to save this information later on.
probably we can use jack metadata :)

Houston4444:
Done the max.
I just adapt the item selection behavior, but I don't understand why my port_group_widget doesn't auto select its concerned connections as it should.

There are maybe changes that you made in canvas port 3 last years that I could have erase. You would know better than me what to do with if this is the case.

falkTX:
Thanks, I will review this soon.
Sonoj is in a few days, need to focus. So at least after that.

Houston4444:
Hi !
Any news for this PR ? Did you take time to try it and take a look ?

falkTX:
Sorta, it is mostly okay, I think my idea was to rename things a little to fit better into existing code style, but it has been months since I looked at this so I might be wrong.

Something I will try to add for v2.4, but wont make it to v2.3

Houston4444:
Ok thanks. I saw there are now conflicts, but all seems to be very simple.

PR #943, opened 2019-10-23
Wheel knobs

Houston4444:
That was finally quite easy to do !

Make mouse wheel works on knobs, with little steps when Ctrl is Pressed.
Double click makes knob turn to the mouse cursor position (if you don't want this, simply remove the doubleClickEvent method, it doesn't needs extra stuff).

Notes: something better could be done with the captation of Ctrl press/release. Ctrl doesn't affects the knob step if a dialog window has focus. But that's a very little thing.

### DIFF
--- a/source/frontend/carla_host.py
+++ b/source/frontend/carla_host.py
@@ -131,9 +131,9 @@ class HostWindow(QMainWindow):
 
         # ----------------------------------------------------------------------------------------------------
         # Internal stuff
-        
+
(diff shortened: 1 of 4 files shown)


falkTX:
Thanks for updated PR.
I really dislike that global ctrl handling... regarding scalepoints, they should be sorted by default. If not, then should be fixed in the engine. we do not want to sort scalepoints every single time.

PR #982, opened 2019-12-19
Enable IRC notifications only

redtide:


falkTX:
Thanks, let's see if this works :)

PR #1073, opened 2020-04-13
discovery: print category for vst plugins

alex-tee:


source/discovery/carla-discovery.cpp
```diff
@@ -1122,6 +1123,34 @@ static void do_vst_check(lib_t& libHandle, const char* const filename, const boo
         else
             cVendor.clear();
 
+        // get category
+        const intptr_t effCategory = effect->dispatcher(effect, effGetPlugCategory, 0, 0, NULL, 0.0f);
+        category = PLUGIN_CATEGORY_NONE;
```
falkTX, line 1128: this should not be here, as you are setting the variable again right after in the switch case.
please add a default case in the switch, and set this there instead.

source/discovery/carla-discovery.cpp
```diff
@@ -1122,6 +1123,34 @@ static void do_vst_check(lib_t& libHandle, const char* const filename, const boo
...
+            break;
+        case kPlugCategRestoration:
+            category = PLUGIN_CATEGORY_UTILITY;
+            break;
+        case kPlugCategGenerator:
+            category = PLUGIN_CATEGORY_SYNTH;
+            break;
+        }
+
+        if (effect->flags & effFlagsIsSynth)
```
falkTX, line 1151: this should only happen if category is none (the default switch case)

falkTX:
Thanks!

PR #1089, opened 2020-04-18
Fix incorrect base conversion for midi CC numbers

### LINKED ISSUE #1088: Incorrect MIDI CC port base conversion
It seems there is a base conversion issue when sending CC 0x10 Carla ignores it for parameter automation unless you select Midi Control 10 (0xA).

This can also be seen as when 0x10 is selected the grey text at the top of the context menu states it's on control 22 and not 16 decimal.

gnif:
Fixes #1088


falkTX:
Tested and seems correct indeed. thanks!

PR #1094, opened 2020-04-20
Prepare for native VST3 implementation / JUCE LINUX VST3 support

terencode:


source/backend/plugin/CarlaPluginVST3.cpp
```diff
@@ -30,7 +30,6 @@ CarlaPlugin* CarlaPlugin::newVST3(const Initializer& init)
 #ifdef USING_JUCE
     return newJuce(init, "VST3");
 #else
-    init.engine->setLastError("VST3 support not available");
```
falkTX, line 33: this line should not be removed

source/backend/utils/Information.cpp
```diff
@@ -65,10 +65,10 @@ const char* carla_get_complete_license_text()
         "<li>LADSPA plugin support</li>"
         "<li>DSSI plugin support</li>"
         "<li>LV2 plugin support</li>"
-#if defined(USING_JUCE) && (defined(CARLA_OS_MAC) || defined(CARLA_OS_WIN))
+#if defined(USING_JUCE) && (defined(CARLA_OS_MAC) || defined(CARLA_OS_WIN) || defined(CARLA_OS_LINUX))
```
falkTX, line 68: we can remove the OS check here and only check if juce is being used. since it only works for these OSes anyway

source/backend/utils/Information.cpp
```diff
@@ -65,10 +65,10 @@ const char* carla_get_complete_license_text()
...
         "<li>VST2/3 plugin support (using Juce)</li>"
 #else
-        "<li>VST2 plugin support (using VeSTige header by Javier Serrano Polo)</li>"
+        "<li>VST2/3 plugin support (using VeSTige header by Javier Serrano Polo for VST2)</li>"
```
falkTX, line 71: This is not correct, the VeSTige header is only for VST2, it cannot do VST3

falkTX: no nvm, this is correct! just that it would be better as a separate line

source/backend/utils/Information.cpp
```diff
@@ -199,7 +199,7 @@ const char* const* carla_get_supported_features()
 #endif
 #ifdef USING_JUCE
         "juce",
-# if defined(CARLA_OS_MAC) || defined(CARLA_OS_WIN)
+# if defined(CARLA_OS_MAC) || defined(CARLA_OS_WIN) || defined(CARLA_OS_LINUX)
```
falkTX, line 202: same here, we can just remove the OS check completely

source/discovery/carla-discovery.cpp
```diff
@@ -1676,7 +1676,7 @@ int main(int argc, char* argv[])
 #ifdef USE_JUCE_PROCESSORS
         do_juce_check(filename, "VST3", doInit);
 #else
-        DISCOVERY_OUT("error", "VST3 support not available");
```
falkTX, line 1679: this does not work, needs to be reverted. my native vst dicovery is only for vst2

source/frontend/carla_database.py
```diff
@@ -472,8 +472,8 @@ def run(self):
         # Increase count by the number of externally discoverable plugin types
         if self.fCheckNative:
             self.fCurCount += pluginCount
-            # MacOS and Windows are the only VST3 supported OSes
-            if self.fCheckVST3 and not (MACOS or WINDOWS):
+            # Linux MacOS and Windows are the only VST3 supported OSes
```
falkTX, line 475: Nitpick, but missing comma after Linux

PR #1107, opened 2020-04-27
fix #1102 rpath specification in pc files

### LINKED ISSUE #1102: carla lib fails to link on macos
`ld: unknown option: -rpath=/usr/local/lib/carla`

replacing `=` with `,` in the pc files fixed it

alex-tee:


falkTX:
Can you confirm this works on other systems too?

alex-tee:
works on GNU/Linux, just linked against this

it's a PITA to check on windows but mingw tools afaik are the same GNU tools i use on GNU/Linux so it's probably fine there too

falkTX:
ok, thanks for testing

PR #1114, opened 2020-05-03
Update Travis syntax

edumco:
This PR removes old syntax in favor of a more recent one


falkTX:
Thanks!
I have been not been paying attention to CI changes, usually as long as it keeps working it is fine by me. Assuming it works, this prevents future pain when the deprecated stuff is removed, so it is nice to have.

PR #1122, opened 2020-05-09
Fix MinGW JUCE DEBUG builds

terencode:
Compilation was failing with:

```
/usr/lib/gcc/x86_64-w64-mingw32/9.3.0/../../../../x86_64-w64-mingw32/bin/as: ../../../build/juce_gui_basics/Debug/juce_gui_basics.cpp.win64.o: too many sections
{standard input}: Fatal error: can't close ../../../build/juce_gui_basics/Debug/juce_gui_basics.cpp.win64.o: file too big
```

Solution is to allow bigger object files with the `-mbig-obj` GNU assembler
flag

Source: https://digitalkarabela.com/mingw-w64-how-to-fix-file-too-big-too-many-sections/

### DIFF
--- a/source/modules/juce_audio_processors/Makefile
+++ b/source/modules/juce_audio_processors/Makefile
@@ -18,7 +18,7 @@ else
 # needed by vst3
 BUILD_CXX_FLAGS += -w
 ifeq ($(DEBUG),true)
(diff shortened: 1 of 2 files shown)


falkTX:
I noticed some old checks still in the makefile, I will commit a simplified version of this, thanks!

PR #1125, opened 2020-05-10
Fix JUCE compilation on Linux with DEBUG=true

terencode:
GCC does not recognize the `-mbig-obj` MinGW flag, it should only be
added if DEBUG is true AND we are crosscompiling.


falkTX:
oh, did not realize that.
still, the PR is not exactly correct, please do not add indentation in the makefiles, it is not valid syntax.
also the juce_gui_basics module needs this too

terencode:
Both issues fixed.

falkTX:
you still have indentation in the makefiles.
again, this is not valid

terencode:
Sorry I thought you said "do not add tabs"...fixed now.

falkTX:
cool, thanks.
this ought to trigger CI, but it didnt, so I dont have a way to check if it passes or not :(
anyway will merge.

PR #1160, opened 2020-06-01
Fully Scalable UI Part 1

terencode:
Code preparation for scalable LEDs, Dials and Buttons

_This is part 1 on a serie of PRs addressing some blurry UI elements on HiDPI displays._

### DIFF
--- a/resources/resources.qrc
+++ b/resources/resources.qrc
@@ -44,7 +44,7 @@
 
         <file>scalable/carla.svg</file>
         <file>scalable/carla-control.svg</file>
-        
+
         <file>scalable/folder.svgz</file>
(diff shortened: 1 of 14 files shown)


falkTX:
this is not really clean, since it mention files not present in this PR.
But i will take it, otherwise we are here forever.
Please push the svg files references in this PR into the next one, and only those files.
Then we can see about the rest

terencode:
> this is not really clean, since it mention files not present in this PR.

Yep forgot about it when moving commits around, I could have changed it but thanks for merging already.
> Please push the svg files references in this PR into the next one, and only those files.
Then we can see about the rest

This is what I had in mind, done.

PR #1162, opened 2020-06-01
Fully Scalable UI Part 3

terencode:
Add non-default rack LEDs, dials and buttons SVGs

### To Do:

- [x] Calf Leds
- [x] Calf buttons
- [ ] Custom dials

_This is part 3 on a serie of PRs addressing some blurry UI elements on HiDPI displays._


falkTX:
any news on this? reconsidering allowing calf stuff now.
even though our resources are old, they still work. so I am okay with using them.
would be nice to delete the old bitmap files while adding the new svg ones.

PR #1314, opened 2020-12-15
Use freedesktop theme icons on Linux with svg ones as fallback

redtide:


falkTX:
Thanks, this is a good start.
But I am a little worried about some things:
1 - it needs to be optional, opt-in, because by default it is not possible to know if the icon theme is going to clash with the custom black theme
2 - verify that works on win/mac without side-effects (the hardcoded/built-in icons should always be used there, never system ones)
3 - verify that HiDPI still works. I dont have such a screen myself, but we can setup env vars for qt to render at 2x or even more scaling, which allows us to test if icons look sharp or blurry.

redtide:
1 - I'm not sure what you mean with "clash with the custom theme", I think it was the theme I used before switching to the native in my system and the icons seems to be the same.
2 - I'm building it on Windows right now, just fighting against msys2 stuff :) system icons? Theme icons should work only on Linux, and fallback to the original resource ones on other systems. Even on Linux if an icon is missing it fall back with the resource one (see the link in PM on IRC).
3 - I'm afraid that depends on the icon theme installed in the user system, so it could be OK or look crap... I suppose that this could be a reason for an option to disable them.

falkTX:
1. I mean the user might have a bright/light theme with dark icons, which would clash with carla built-in default dark theme. so I rather have this opt-in
2. all fine if mac/win system icons are never used, just need to verify that is the case
3. yes, depends on the theme. but also you have some parts of the code with `pixmap(22,22)` which I am not sure works well in High DPI scenarios, testing needed

redtide:
As reminder: needs a getIcon() function that can return both svg and pngs to facilitate hidpi with scaled icons, add an UI option to disable this capability on Linux where scaled icons cannot be used or not fitting well with dark themes, then mac/win and linux hidpi tests.

redtide:
WIP in develop, closing. Thank you!

PR #1383, opened 2021-03-21
FreeBSD patch in AppConfig.h

yurivict:


falkTX:
Does this actually work though?
There is a filter for linux and juce in the makefiles, which this patch does not address

yurivict:
I added/removed LV2 plugins and it worked.

What function should I test?

falkTX:
vst3 plugins. that is the only reason juce is used in carla under linux.
if you do not have any vst3 plugins in freebsd, this patch is pointless (for now)

yurivict:
Many vst plugins are available from FreeBSD ports, ex. ```lib/vst/ZamCompX2-vst.so``` or ```lib/vst/midipressuretocc-vst.so```. Not sure if they are vst2 or vst3.

falkTX:
those are vst2, not vst3.
try building the vst3sdk first, it has example plugins.

falkTX:
This might be incomplete, but makes the build on BSD pass, so let's use it.
Thanks!

Do note that I will replace the juce vst3 hosting with a custom implementation, so then linux (and bsd) builds won't be using juce anymore.
but that will take a while

PR #1417, opened 2021-05-08
Fix Windows cross compilation

edumco:
Fix win32 and win64 builds by:
- adding a missing pub key
- update trusty to bionic (some missing packages in trusty)
- migrate from mingw32/mingw64 packages to mingw-w64


falkTX:
travis is long gone, RIP travis.

PR #1426, opened 2021-05-21
compile python files to reduce program startup time

Houston4444:
Hi.
The idea is to compile .pyc files in __pycache__ at make install.

This reduce the startup duration of carla. When a python file is imported, python compiles this file in __pycache__/file.pyc . If this file already exists and matches perfecty with the python file version, then, python uses the compiled file. If we don't install theses compiled files, compilation is done at each startup because the dir is read only.

for comparaison, see the return without __pycache__ of `time carla --version`

```
time carla --version
Using Carla version 2.3.0
  Python version: 3.7.3
  Qt version:     5.11.3
  PyQt version:   5.11.3
  Binary dir:     /usr/local/lib/carla
  Resources dir:  /usr/local/share/carla/resources

real    0m0,707s
user    0m0,605s
sys     0m0,082s
```

and with __pycache__:
```
Using Carla version 2.3.0
  Python version: 3.7.3
  Qt version:     5.11.3
  PyQt version:   5.11.3
  Binary dir:     /usr/local/lib/carla
  Resources dir:  /usr/local/share/carla/resources

real    0m0,426s
user    0m0,360s
sys     0m0,049s
```

Note that the compilation is done at install, and not at make, else when files are moved, compiled files are not considered as matching with python files. It means that python checks if it matches with absolute paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ MODULEDIR := $(CURDIR)/build/modules/Release
 endif
 
 VERSION   := 2.3.0
-
+PYTHON    := python3
 # ---------------------------------------------------------------------------------------------------------------------
 
 all: backend discovery bridges-plugin bridges-ui frontend interposer libjack plugin theme
@@ -608,6 +608,8 @@ endif
 		bin/resources/carla-plugin-patchbay \
 		bin/resources/*-ui \
 		$(DESTDIR)$(DATADIR)/carla/resources
+		
+	
 endif # CPPMODE
 
 ifeq ($(HAVE_THEME),true)
@@ -767,6 +769,10 @@ ifeq ($(CAN_GENERATE_LV2_TTL),true)
 endif
 endif
 
+	# -------------------------------------------------------------------------------------------------------------
+	# compile all python .pyc files to __pycache__ for faster application start
+	$(PYTHON) -m compileall $(DATADIR)/carla
+	
 # ---------------------------------------------------------------------------------------------------------------------
 
 ifneq ($(EXTERNAL_PLUGINS),true)


Houston4444:
I think the checks fails because this PR makes python3 as a new build dependency, but I don't know how to fix this.

edumco:
Try add python to the .travis.yml

python:
  - "3.8"

falkTX:
Travis is dead. I am in the process of updating carla to github actions, not finished.

edumco:
@falkTX great to hear about Actions, 

How this python dependency could be solved?

Can we make it call any python version?

falkTX:
it needs to be python3 for sure, but detection for python is needed. so such block would be inside a check for frontend

falkTX:
note that this is purely a startup optimization.
runtime remains unaffected afaik.

edumco:
@falkTX thats why i asked for some benchmark, so you could evaluate the ROI

falkTX:
Isnt that what @Houston4444 did on the very first post?

anyway, this is a very minor optimization and one that can easily break due to packaging differences in distros.
the python3 executable is not guaranteed to be the same version used for pyuic/pyrcc tools

Houston4444:
> anyway, this is a very minor optimization and one that can easily break due to packaging differences in distros.
the python3 executable is not guaranteed to be the same version used for pyuic/pyrcc tools

Minor, yes, but appreciable. There is probably a way to use it if it works and not care if it doesn't.

heruki:
@Houston4444 I think it makes sense to have it in the github workflows... it's already there by now

PR #1435, opened 2021-06-04
Add missing separator in SF2 search path

Kirtai:
The SF2 search path is missing a separator which results in a config file entry like 

`SF2=/home/faith/.sounds/sf2/home/faith/.sounds/sf3, /usr/share/soundfonts, /usr/share/sounds/sf2, /usr/share/sounds/sf3
`
by default. Note the missing comma between the first "sf2" and the following "/home"

This pull simply adds the missing ":" in the source.


falkTX:
Yeah this makes sense. a copy&paste typo.
Thanks

PR #1455, opened 2021-07-28
Set flag `windowAppearsOnTaskbar` on plugin window

jpcima:
There was a curious regression introduced in a79c192, which makes the plugin window behave badly, when it's run under `carla-single`.
It would not appear in the window list, and it would cause the window ordering and focus to act erratically. (at least in XFCE desktop)

### DIFF
--- a/source/utils/JucePluginWindow.hpp
+++ b/source/utils/JucePluginWindow.hpp
@@ -106,7 +106,7 @@ protected:
 
     int getDesktopWindowStyleFlags() const override
     {
-        return ComponentPeer::windowHasDropShadow | ComponentPeer::windowHasTitleBar | ComponentPeer::windowHasCloseButton;
+        return ComponentPeer::windowAppearsOnTaskbar | ComponentPeer::windowHasDropShadow | ComponentPeer::windowHasTitleBar | ComponentPeer::windowHasCloseButton;
     }
 
 #ifndef CARLA_OS_LINUX


falkTX:
Right, this should not be there if running carla-single, but it is still useful when running the main carla.
Need to put in place a way to detect this.

jpcima:
Another observation just now
- when a VST3 plugin is ran in carla-single, I try to alt-tab the window to another, it will refuse to let go of the window focus unless patched by this PR.
- a VST2 plugin still refuses to let go of the focus, regardless if patched or not.

falkTX:
On linux, VST2 plugins are not handled by JUCE, so makes sense that this patch keeps the same behaviour

jpcima:
Thanks for the hint.
Now I've established that it's this commit which created the VST2 version of the issue. f7320fe

falkTX:
And you are using latest git?
I fixed something regarding that new code couple days ago.

Still, seems weird for this to happen, which plugin does not give away focus?

jpcima:
I use a git version dated from today, that which this PR is based on.
Also the problem applies to all VST2, DPF and JUCE ones alike.
In VST3 I've tried sfizz only.

falkTX:
Now handled in latest git, specifically 1ff775e7269bd57ccd3e5bb089fed58f89f197e7
The VST2 focus issue should be fixed in e32379407694e867057e6bf26e84314482590b92

testing would be welcome.
though the PR itself can be closed now, as carla now has a way to detect such things.

PR #1468, opened 2021-09-04
Install header CarlaPluginPtr.hpp

jpcima:
It's not possible to develop using carla-host-plugin without this header.


falkTX:
That makes sense, thanks!

PR #1483, opened 2021-10-06
Add disconnect-all-of-group menu item to patchbay

sebageek:
If one group has multiple connections to another group an extra menu
item is added to its disconnect context menu, which allows the user to
disconnect it all at once. The entry is added at the bottom of the
disconnect menu in the form of "Groupname (all x ports)". This comes in
very handy when the user wants to disconnect a soundcard from their
current application.

PortContextMenuDisconnect() now also accepts a list via
setData()/data(), so we can disconnect multiple ports at once.

---

Heyhey, first PR on this project, hope this feature is interesting to you!
I'm a bit unsure about how the patchcanvas module is maintained. I know it is at least used in Catia. Would we need to port this change over? Especially as I augment the functionality of `PortContextMenuDisconnect()`.

### DIFF
--- a/source/frontend/patchcanvas/canvasbox.py
+++ b/source/frontend/patchcanvas/canvasbox.py
@@ -502,10 +502,31 @@ class CanvasBox(QGraphicsObject):
                     conn_list_ids.append(tmp_conn_id)
 
         if len(conn_list) > 0:
+            groups = {}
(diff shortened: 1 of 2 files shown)


sebageek:
Rebase to current main + added a separator between the normal entries and the "cut a complete group" entries.

PR #1485, opened 2021-10-09
Replace -lpthread with -pthread

Xeonacid:
According to [Stack Overflow](https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling) and [GCC doc](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#Preprocessor-Options) (search for `pthread`), -pthread is the correct way to use thread.

This fix build error on riscv64.
Before:
```text
/usr/bin/ld: ../../build/modules/Release/hylia.a(hylia.cpp.o): in function `.L1897':
hylia.cpp:(.text.hylia_enable+0x20): undefined reference to `__atomic_exchange_1'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:265: ../../bin/libcarla_native-plugin.so] Error 1
```

Tested on x86_64 and riscv64 on Arch Linux.


falkTX:
I have read about this before, but never appeared the occasion where -lpthread was not enough.
But makes sense, thanks!

PR #1495, opened 2021-10-14
JSFX plugin format

jpcima:
JSFX work in progress (2nd attempt)

- [x] storage of serialization data in base64 format
- [x] provide API for File and GFX, to prevents "undefined" compilation errors

To-do
- [x] size correction of MIDI events which should be 2 instead of 3
- [ ] possible MIDI-related bugs


falkTX:
active sense messages should just be ignored by jack I think.
if not try with a midi filter before the jsfx that blocks those

jpcima:
> active sense messages should just be ignored by jack I think.

I'm on pipewire. PW's problem then?
At a lesser rate, it also receives midi clocks.

Issue #1487, opened 2021-10-10
JSFX plugin format


jpcima:
This is a source-based plugin format, which is made by Cockos and part of Reaper.
(Note that JSFX does not stand for JavaScript.)

There are dozens of these plugins shipped in Reaper (see `InstallData/Effects` folder).
Most are rather short source code files, tagged with a free software license on them.

The implementation of JSFX is partially free software.
[EEL2] is the underlying programming language, which is published as part of WDL.
There exists a free software implementation in [jsusfx].

Some notes
- the language would be useful for prototyping all sorts of dsp
- the files are arranged in a multi-level folder hierarchy. one might need to compile them in order to scan them
- the source is expected to compile at runtime to native assembly ; the build system is a little convoluted, which has architecture-dependent code and involves assembly in the nasm syntax.
- the jsusfx implementation is last updated 3 years ago, and the EEL2 is a bit behind; in particular, the later one has support of the arm 64bit processor.

[EEL2]: https://github.com/justinfrankel/WDL/tree/master/WDL/eel2
[jsusfx]: https://github.com/asb2m10/jsusfx/

falkTX:
lets keep the discussion here then.
my main doubt against all this is a sorta "vendor lockin" if we place this in carla. It kinda makes sense for carla to load them since it is really a plugin format in the end, plus the drag&drop would make this handy... but it would be something limited to carla instead of being usable everywhere.

the dynamic nature of these files likely makes it very tricky to have as general plugins though, that part I do not know how to solve.

If carla really ends up supporting jsfx, then I would see it as a motivation to do a carla mini-host, a 1-to-1 wrapper for the things carla is able to load.

how fast is the scanning for these files?

jpcima:
There is the current status at this branch.
Currently it builds the support libraries.
https://github.com/jpcima/Carla/tree/jsfx

jpcima:
Note: there appears to be a mistake in file `jsusfx.cpp`
The following should be modified from `ifdef` to `ifndef`, to let scripts print their messages.
`#ifdef EEL_STRING_STDOUT_WRITE`
In addition, it should preferably use the carla-specific IO functions.

jpcima:
[ ] Other: values don't parse correctly when `LC_NUMERIC != C`
Test case is Reaper `guitar/phaser`, which fails to modulate anything, because of multiplier `0.5` parsed as `0`.

falkTX:
[...]

PR #1503, opened 2021-10-24
jsfx: identification by relative path

jpcima:
#1487

This work would allow JSFX to be identified by their relative file ID.
This relative file ID is kept in the `label` field.

The JSFX unit will be identified by the pair of these two strings: root path, file ID
Their concatenation produces the full path.
A file ID is normalized to forward-slash separators.

The CarlaJsfxPlugin is edited such that it can load by either `filename` or `label`.

~~Draft because it's not tried yet.~~


jpcima:
Changed the plugin name to use the `desc` field of jsfx files, because it's what Reaper does.
This gives more descriptive names.

PR rebased, nothing more to add

PR #1508, opened 2021-11-01
JSFX misc

jpcima:
This corrects a pair of issues in the JSFX implementation. #1487

- [x] the source should be compiled once only, so that I/O ports are stable
  Not recompiling might cause the plugin to be not clean as new after just calling an `@init` (by `fEffect->prepare`); this can be left to the script programmer after all.

- [x] plugin `getLabel` would not return the proper value, it was a leftover from before, when using full paths as label rather than a relative path fragment. Replaced the path `fFilename` with the 2-part identifier. (`CarlaJsfxUnit`)

Ok for merging.

### DIFF
--- a/source/backend/plugin/CarlaPluginJSFX.cpp
+++ b/source/backend/plugin/CarlaPluginJSFX.cpp
@@ -283,18 +283,6 @@ public:
 
         // ---------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


falkTX:
Looks good, thanks

PR #1529, opened 2021-11-15
jsfx: cleaned up implementation

jpcima:
This is a remake of JSFX work.

Most notably, it removes jsusfx and substitutes it with a newly written library.
This library has a cleaner interface and is better tested.

Also, the work was rearranged into a pair of commits frontend/backend.

It's working; there's more about implementation status here: https://github.com/jpcima/ysfx/issues


jpcima:
Problem: sliders which have unspecified ranges prevent the plugin from running correctly
as an example, `slider6:0,FFT Size` from `guitar/amp-model-dual`, a convolution plugin

The engine will crash with `Parameter 'FFT Size' too high! - 256.000000/1.000000`.
The ideal for this parameters is a text box which would accept any value.

falkTX:
@jpcima this is ready to merge, correct?

I would go without gui or these weird parameters for an initial version, just so we can get things rolling.

jpcima:
Not yet sorry, a few fixes went in the library since the PR, thanks to a couple of reports.

Reminder to self: library
- [ ] add the function calls to register default image/sound loaders, so that convolution plugins will work
- [ ] add latency support to library (which might be dynamic, and also may affect only a subset of outputs)
- [ ] misc language features: gmem, hints, ...

also I notice a lot of eel2 has landed in WDL upstream, need to catch up with that.

jpcima:
btw Cockos is really wanting to nuke the float 32bit support
https://github.com/justinfrankel/WDL/commit/058298340fae37fd670f158ad5c620549671c40c

falkTX:
> Not yet sorry, a few fixes went in the library since the PR, thanks to a couple of reports.

ok, was just wondering.

I am thinking of already adding the definitions so 3rd party code that uses carla as library can already have these JSFX definitions available to use. It is a bit annoying to ifdef based on version, like in ildaeil and cardinal.



> btw Cockos is really wanting to nuke the float 32bit support [justinfrankel/WDL@0582983](https://github.com/justinfrankel/WDL/commit/058298340fae37fd670f158ad5c620549671c40c)

that is quite sad to see. double usage has a performance impact on embed machines :(
is your library affected by this? or can you still keep float support regardless?

jpcima:
> that is quite sad to see. double usage has a performance impact on embed machines :(
is your library affected by this? or can you still keep float support regardless?

It's not too difficult to add these changes back in, of course, but the more patches pile up and the harder it is to follow the updates. One still would have to write these 32bit assembly parts to enable fast code.

There exists another consideration: jsfx is one of these languages, like js and php, that represents integers as floats. That means a lot of (int) casting implied and it raises the question about the performance of doing this repeatedly (esp for array indexing)

A related thing, some API are relying on bitmasking (eg. a mask of sliders to automate)
and single-float can encode successive integers only up to 24 bits, so that might possibly create trouble as well.

jpcima:
An update of JSFX support
- updated library and WDL
- linked `@gfx` support code and platform dependencies (support UI later)
- added latency support. Note: it could report latency updates after `run`, but it follows LV2 which doesn't

Problem remaining
- still some build problems, will fix shortly
- non-JSFX buddy files getting discovered as plugins (.RPL files, which are preset libraries)
- display output params force-clamped to (0;1)

jpcima:
One remark, maybe we'd rather wait that I figure out how the reaper .rpl format works?
This can enable support for programs, but also I can RE the base64 format used there, and I can then implement the identical one in Carla. Not that it's absolutely vital to do so, but for a perspective of future compatibility.

falkTX:
there is no need to rush.
since the idea seems to be to not merge this right now for next quarterly release, we can wait until that passes and merge aferwards.

I do want to get some of the definitions in place already though, in order to simplify some external code (mainly in ildaeil).
This will generate some small conflicts

source/backend/engine/CarlaEngine.cpp
```diff
@@ -596,6 +596,7 @@ bool CarlaEngine::addPlugin(const BinaryType btype,
                            && ptype != PLUGIN_GIG
                            && ptype != PLUGIN_SF2
                            && ptype != PLUGIN_SFZ
+                           && ptype != PLUGIN_JSFX
```
falkTX, line 599: any special reason for putting this line and making jsfx not able to be bridged?
Since they are external files with arbitrary code, I think allowing them as bridge is generally desired.

jpcima: Sure, bridging may have some benefits. Running jsfx compiled code is safe against crashing, but is not safe against things like infinite loops.
A thing of note is also that jsfx can't separate DSP and UI, and I don't understand too much yet how this connects with Carla's bridge concepts overall.

jpcima:
A few things about the current update
- rebase on definitions from the main branch
- the support library has received ability to load reaper presets; no Carla implementation yet, this probably needs some help. An example plugin is Atlantis Reverb from here: https://github.com/geraintluff/jsfx/
- jsfx bridge disabled (for now), I don't manage to deal with makefile complexity to make it build; if you don't mind, please check it whenever you have a moment

falkTX:
@jpcima I am merging this now, will deal with cleanup and adjusting to carla code style (whatever that is, I cant really decide myself)
I want to split the carla discovery cpp file, as it got too big, so this is a good time.

PR #1535, opened 2021-11-17
add app metainfo

Bleuzen:
Hi, so this adds a xml file with some metadata about Carla. This is required since I currently try to package Carla as flatpak:
https://github.com/flathub/flathub/pull/2645
and they prefer to have the appdata in the upstream repo.
Also we can discuss possible changes. For example the screenshots and app id.


falkTX:
Fine by me, but this should be auto generated (for the version) no?
Also should go into the data/ subdir, not on the root of the repo.
Does it need to get installed somewhere?

falkTX:
There are several issues with the flathub entry btw.. forcing X11 mode being the biggest one, afaik Carla works fine under wayland too. The X11 stuff goes through XWayland just fine, Carla does not use Qt methods to achieve this (only for VST plugin version, but we can fix that part)

Bleuzen:
> but this should be auto generated (for the version) no

This is up to you. For example you could even use a template and generate this file during build.



> Also should go into the data/ subdir, not on the root of the repo.

Okay, I will move it.

> Does it need to get installed somewhere?

Yes. Should land in the share/appdata directory.
Flatpak:
`/app/share/appdata/com.github.falkTX.Carla.appdata.xml`
Native:
`/usr/share/appdata/com.github.falkTX.Carla.appdata.xml`

Bleuzen:
> There are several issues with the flathub entry btw.. forcing X11 mode being the biggest one,

I removed the QT_QPA_PLATFORM env var now.

PR #1541, opened 2021-11-27
fix default sf2/3 path

Bleuzen:


falkTX:
Good catch, will merge.

PR #1542, opened 2021-11-27
detect pipewire jack backend for better defaults

Bleuzen:
These days there is pipewire, which provides backends for PulseAudio, ALSA and JACK. Usually, if one is using pipewire as a replacement for the former three, they want to run Carla with the JACK backend mode since it is the most compatible and allows managing pipewire connections.

If the pipewire jack backend is installed can be detected by checking if `/usr/bin/pw-jack` exists, so I added it to the check.


falkTX:
That makes sense, thanks!

Issue #1549, opened 2021-12-13
Python3.10.1/PyQT - draw elements errors


BlueMax:
I have several Python errors in Carlas log since yesterday when Python was updated to 3.10.1.
Most plugin elements in "Rack" tab are missing and Carla hangs at drawing the first plugin. 
Downgrading to Python 3.9.9 leads to Carla  'PyQT not found'. Probably needs more Python deps to be downgraded as well. 
There were other updates like Pipewire in the meantime so i'm not 100% sure about Python. 'lib32-pipewire' and 'lib32-pipewire-jack' are not yet updated/available but i think these are of minor relevance.

Can anyone confirm this bug or is it just me?

Carla log:
https://gist.github.com/BlueMax/1160912968cda3f95c0e3854c18021c4/raw/8eed12017ee7ee81d18ec3b641654c68fd0379af/gistfile1.txt

This is my patchbay/project:
[carla-patchbay.png](https://user-images.githubusercontent.com/1403092/145848172-b2655473-0f02-4edb-a8f7-a57f5fe7b81f.png)

```
Arch Linux 5.15.7
local/carla-git 1:2.4.1.r27.g8c8efd2b1-1

Carla 2.4.1 started, status:
  Python version: 3.10.1
  Qt version:     5.15.2
  PyQt version:   5.15.6
  Binary dir:     /usr/lib/carla
  Resources dir:  /usr/share/carla/resources
Frontend pixel ratio is 1.0
libjack.so.0 loaded successfully!
```
```
local/lib32-pipewire 1:0.3.40-1
    Low-latency audio/video router and processor - 32-bit client library
local/lib32-pipewire-jack 1:0.3.40-1
    Low-latency audio/video router and processor - 32-bit client library - JACK support
local/pipewire 1:0.3.41-1
    Low-latency audio/video router and processor
local/pipewire-alsa 1:0.3.41-1
    Low-latency audio/video router and processor - ALSA configuration
local/pipewire-jack 1:0.3.41-1
    Low-latency audio/video router and processor - JACK support
local/pipewire-jack-dropin 3-2
    Use pipewire as drop-in replacement for JACK
local/pipewire-media-session 1:0.4.1-1
    Example session manager for PipeWire
local/pipewire-pulse 1:0.3.41-1
    Low-latency audio/video router and processor - PulseAudio replacement
```

falkTX:
This is the same as #1548

I do not have such new systems, so I cannot reproduce.
Patches are welcome.

[...]

PR #1550, opened 2021-12-14
Fixed unexpected type errors #1549

amazingidiot:
Hello,
I traced the drawing errors and was able to get the meters and knobs working again. Plugins show up in the rack as intended.

#1549

I tested it successfully on my machine. I guess it should work on older python versions, too.

Best regards,
Oliver

### DIFF
--- a/source/frontend/widgets/digitalpeakmeter.py
+++ b/source/frontend/widgets/digitalpeakmeter.py
@@ -21,7 +21,7 @@
 
 from math import sqrt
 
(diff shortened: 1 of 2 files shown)


falkTX:
Oh nice, you even went the extra mile and used the proper QLine/RectF types instead of converting everything to int.
Thanks a lot!

amazingidiot:
Cool!

Happy to help :+1:

Issue #1552, opened 2021-12-19
modules in the patchbay reset positions when loading project file.


Skrappjaw:
Every time I load up Carla the positions are reset or random. Some in the top left corner or bottom left. Also happens when selecting/moving multiple modules at the same time.
![Carla-Pathcbay](https://user-images.githubusercontent.com/21233073/146680468-ba3f72d9-350e-4764-8d5d-62c0abe4007a.png)

Also, this is how i startup jack/pulse and config my audio bridges....

>#!/bin/bash
killall -9 carla
killall -9 jackdbus
killall -9 pulseaudio
sleep 5
jack_control start & 
sleep 1
pulseaudio --start
cadence &
sleep 2
pacmd unload-module module-jack-sink
pacmd unload-module module-jack-source
sleep 2
#pulseaudio bridges | in / out
pacmd load-module module-jack-source channels=2 source_name=voip-in client_name=voip-in connect=false
pacmd load-module module-jack-sink channels=2 sink_name=voip-out client_name=voip-out connect=false
pacmd load-module module-jack-sink channels=2 sink_name=games-out client_name=games-out connect=false
pacmd load-module module-jack-source channels=2 source_name=media-in client_name=media-in connect=false
pacmd load-module module-jack-sink channels=2 sink_name=media-out client_name=media-out connect=false
sleep 2
#a2jmidid -e &
sleep 5
#carla &
carla ~/carla.setup.carxp &
sleep 3
wall "Audio systems up and running"
sleep 1
exit 0

I've tried starting carla with and without the project file being opened via the script. 

OS- Garuda Linux on Zen Kernel.

mxmilkiib:
Exactly what versions of JACK and Carla are you using? I think at least JACK might be out of date.

Skrappjaw:
Carla - 2.4.1
jack2 - 1.9.19

All3xJ:
I have the same issue with Carla 0.9. The patch https://github.com/falkTX/Carla/pull/1554 (so https://github.com/All3xJ/Catia/tree/patch-1 under Catia) does not fix the bug for me.

All3xJ:
[...]

PR #1554, opened 2021-12-22
Fix type error in signal emission

StefansM:
The `CanvasBox.positionChanged` signal expects to receive `x` and `y` as
`int`s, but they were being passed as `float`s, which were mangled at the
receiving signal handler.

This manifested as huge `x` and `y` values for canvas boxes, and when
saving and loading projects all boxes would hit the edge of the canvas
and get placed at the extreme edge.

### DIFF
--- a/source/frontend/patchcanvas/canvasbox.py
+++ b/source/frontend/patchcanvas/canvasbox.py
@@ -411,7 +411,7 @@ class CanvasBox(QGraphicsObject):
             connection.widget.setZValue(z_value)
 
     def triggerSignalPositionChanged(self):
-        self.positionChanged.emit(self.m_group_id, self.m_split, self.x(), self.y())
+        self.positionChanged.emit(self.m_group_id, self.m_split, int(self.x()), int(self.y()))
         self.m_will_signal_pos_change = False
 
     @pyqtSlot()


StefansM:
I don't know much about QT, so my terminology might be wrong - feel free to adjust the PR message before merging!

StefansM:
I believe this addresses #1552.

All3xJ:
This probably fixes this issue: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1945
that I thought was a pipewire problem, but it's the Catia's bug.

All3xJ:
when will this new version go out?

mxmilkiib:
Out for yourself or someone else? What distro are you on? You could clone the forked repo, checkout the branch and make n make install, or if Arch based, copy n edit the AUR PKGBUILD to point to the other repo+branch and build that way, though maybe the PR will be approved soon and using a Carla -git package would suit. Thigh you do mention Catia not Carla..

All3xJ:
@mxmilkiib thanks for answer. [I forked Catia's repo](https://github.com/All3xJ/Catia/tree/patch-1) and edited the PKGBUILD (since I have ArchLinux) and installed Catia. But the bug is still there, so I don't think that this fixes the bug also in Carla...

falkTX:
@All3xJ you are confusing carla and catia ;)

All3xJ:
> @All3xJ you are confusing carla and catia ;)

No ahaha I know they are different. Just I have found it weird that two very similar softwares have the same bug at the same time (https://github.com/falkTX/Carla/issues/1552 was reported 14 days ago, which was more or less the same period I encoutered the bug in Catia). So I think they are in some way related. Again, this commit does not help for me in Catia. I also tried downgrading pipewire, but it does not fix. I don't really know what could have caused the issue.

All3xJ:
Ok, I confirm that this does not fix the bug in Carla too (I builded the patched @StefansM version). Apparently, to fix https://github.com/falkTX/Carla/issues/1552 , is enough to downgrade:
python 3.10.1-2 ->3.9.9-1
python-pyqt5 5.15.6-6 -> 5.15.6-2
python-pyqt5-sip 12.9.0-3 ->12.9.0-1

I'm trying it right now

udf:
@All3xJ Note that downgrading python breaks all other python- packages installed from the repos, it's not even a good temporary solution. We still need to figure out why upgrading to 3.10 breaks it.

PR #1555, opened 2021-12-23
Handle unordered add / position change events

StefansM:
Fixes (some cases of) plugins ignoring saved `(x, y)` coordinates when a
project is loaded. The canvas can get notified of a plugin's position
before the plugin is actually added, so when the plugin is actually it
just goes to the default position.

This happens because `PatchbayClientPositionChanged` events can arrive
before `PatchbayClientAdded` events, as seen in debug logs like this:

```
...
PatchCanvas::joinGroup(30)
PatchCanvas::joinGroup(30) - unable to find groups to join
PatchCanvas::setGroupPos(30, 1118, 591, 0, 0)
PatchCanvas::setGroupPos(30, 1118, 591, 0, 0) - unable to find group to reposition
PatchCanvas::addGroup(30, b'Audio to CV', SPLIT_UNDEF, ICON_PLUGIN)
...
```

If this happens, the canvas box will be placed in the default position,
rather than obeying the `PatchbayClientPositionChanged` signal. This PR
keeps track of missed events and then applies them once the object is
actually added.

### DIFF
--- a/source/frontend/patchcanvas/__init__.py
+++ b/source/frontend/patchcanvas/__init__.py
@@ -145,6 +145,9 @@ class Canvas(object):
         self.initial_pos = QPointF(0, 0)
         self.size_rect = QRectF()
 
+        self.pending_group_joins = set()
+        self.pending_position_changes = {}
+
     def callback(self, action, value1, value2, value_str):
         print("Canvas::callback({}, {}, {}, {})".format(action, value1, value2, value_str))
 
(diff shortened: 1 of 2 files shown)


StefansM:
I'm sure that there is a better way to do this, but some quick googling suggests that QT doesn't guarantee the order in which slots get called, so state will need to be aggregated somewhere.

falkTX:
is this happening using pipewire or jack?

StefansM:
Apologies for the slow response. I've only observed this using Jack. I've not tried with any other backends though, so I can't draw any conclusions from that.

PR #1556, opened 2021-12-23
add CV16 variant with 16 audio ins/outs

alex-tee:


alex-tee:
it looks like I only get 8 CV inputs and 8 CV outputs even though I specified 16. not sure if there is some kind of internal limit preventing 16

falkTX:
Yeah I see it, tweak the values in https://github.com/falkTX/Carla/blob/03e98872a844163c5209ecbdb7ec65312c506671/source/backend/engine/CarlaEngineGraph.cpp#L1706 to allow 16 as max.

alex-tee:
can confirm this now works with Geonkick (32 audio outs) and AMS multiphase LFO (16 CV outs)

PR #1586, opened 2022-02-08
Move second image to it's own screenshot tag

FHEK789:
The screenshot tag shouldn't contain more than one image


falkTX:
Sure, I just adapted what I was given.
Thanks!

PR #1644, opened 2022-07-17
Add #if defined(__FreeBSD__) to avoid #include <alloca.h> on FreeBSD

yurivict:


falkTX:
@jpcima this should go into the ysfx project

yurivict:
https://github.com/jpcima/ysfx/pull/62

falkTX:
We should merge this into carla too, will do that now, thanks!

PR #1649, opened 2022-07-20
add riscv64 support

moodyhunter:
Currently, it's difficult for WDL eel to support riscv64 'natively', so we are enabling EEL_TARGET_PORTABLE as a workaround.

### DIFF
--- a/source/Makefile.deps.mk
+++ b/source/Makefile.deps.mk
@@ -81,6 +81,10 @@ ifneq (,$(filter aarch64%,$(TARGET_PROCESSOR)))
 CPU_AARCH64 = true
 CPU_ARM_OR_AARCH64 = true
 endif
+ifneq (,$(filter riscv64%,$(TARGET_PROCESSOR)))
+CPU_RISCV64 = true
+CPU_RISCV32_OR_RISCV64 = true
+endif
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Set PKG_CONFIG (can be overridden by environment variable)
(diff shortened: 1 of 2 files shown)


falkTX:
Maybe a better approach here is to enable the generic mode if cpu is not within an allow list?
Otherwise we will need yet another check if building for a different cpu.

@jpcima is there a list of cpus that ysfx/jsfx has optimizations for?
perhaps it is just i686, x64 and arm64?

moodyhunter:
That makes sense, well, it's up to you guys to decide whether we can set "generic mode" for all other architectures.
I'll update this PR once that's confirmed :)

jpcima:
> perhaps it is just i686, x64 and arm64?

yes that's it exactly
and the arm32 support which is broken at the moment (according to eel makefile "armv6t2 + VFP")

moodyhunter:
Hi, the PR has been updated, now it has a better way to check for those architectures which have "optimised" code.

falkTX:
I kinda forgot about this, and ended up doing something similar but in a slightly different way (somewhat more ugly..)

Please check if it builds for you, thanks!

moodyhunter:
Hi @falkTX, I've tested and the commit https://github.com/falkTX/Carla/commit/c9fc68aaa370088b588ad6aea7af60efbc16530c works fine :)

falkTX:
Cool thanks for confirming. Closing the PR then.

btw, which board are you testing with?

moodyhunter:
[We](https://archriscv.felixc.at) currently have 1) StarFive VisionFive, 2) AllWinner D1 but I was testing on a qemu-system virtual machine.

PR #1689, opened 2022-10-09
fix pkgconfig files having bad includedir/libdir variables

alex-tee:
The pkgconfig files are installed in `$PREFIX/lib` so using `libdir=${prefix}/lib/carla` is confusing.

Most software pkgconfig files seem to match the includedir and libdir variables with the actual includedir and libdir during configuration.

### DIFF
--- a/data/pkgconfig/carla-host-plugin.pc
+++ b/data/pkgconfig/carla-host-plugin.pc
@@ -1,9 +1,10 @@
 prefix=X-PREFIX-X
-libdir=X-LIBDIR-X/carla
-includedir=X-INCLUDEDIR-X/carla
+libdir=X-LIBDIR-X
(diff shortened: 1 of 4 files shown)


data/pkgconfig/carla-host-plugin.pc
```diff
@@ -1,9 +1,10 @@
 prefix=X-PREFIX-X
-libdir=X-LIBDIR-X/carla
-includedir=X-INCLUDEDIR-X/carla
+libdir=X-LIBDIR-X
+includedir=X-INCLUDEDIR-X
+carla_libdir=${libdir}/carla
```
alex-tee, line 4: I am adding a separate variable here because Carla installs binaries there that I want, so having a specific variable to get the path is useful

PR #1692, opened 2022-10-09
fix make install (missing pluginlist dir)

alex-tee:
latest main branch fails during `make install` without this


falkTX:
thanks, I forgot about that one

PR #1724, opened 2022-12-30
missing </li> code

trebmuh:
isn't?


falkTX:
Seems like it indeed, but you have a typo in `...dov"</li>"` the previously final quote needs to be removed so it becomes `dov</li>"`

trebmuh:
fixed

falkTX:
cool, thanks!

PR #1757, opened 2023-03-18
Fix build on mac m1 

nikage:


falkTX:
Seems good, anything else needed here?

nikage:
> Seems good, anything else needed here?

No, it's fine. I just thought a description i required.

PR #1780, opened 2023-06-01
fix LADISH support

nedko:
Menu items are useful in LADISH sessions,
when Carla not ran in NSM mode


nedko:
I give up (see IRC) there is no point in bureaucracy when faced discrimination "IF LADISH THEN DONT WORK"

PR #1805, opened 2023-09-05
Fix build with clang on linux

vimproved:
Clang does not support the -fno-gnu-unique argument. Previously this was not added if BSD was true (presumably to work around FreeBSD support) but there is a CLANG variable serving that purpose. Use that instead.


vimproved:
@falkTX I modified this PR to add a clang CI, and also to automatically detect compiler based on `--version` output.

PR #1806, opened 2023-09-05
Fix build with musl libc 1.2.4

vimproved:
Musl 1.2.4 gated the LFS64 compatibility shims behind the _LARGEFILE64_SOURCE macro, and they will be removed entirely in musl 1.2.5. Instead, always use the normal stat function instead of stat64, and set -D_FILE_OFFSET_BITS=64, which ensures that functions will always be 64-bits on 32-bit glibc platforms (and does nothing on other platforms).

### DIFF
--- a/source/modules/water/files/File.cpp
+++ b/source/modules/water/files/File.cpp
@@ -1265,7 +1265,7 @@ private:
 //=====================================================================================================================
 namespace
 {
-   #ifdef CARLA_OS_LINUX
+   #ifdef __GLIBC__
     typedef struct stat64 water_statStruct;
     #define WATER_STAT    stat64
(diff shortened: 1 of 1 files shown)


vimproved:
> I believe this breaks mingw and wasm builds, it shouldn't be enabled by default. Lets see if the CI jobs fail

Oh alright, I don't usually work on software with mingw/wasm builds, sorry. Will fix ASAP.

falkTX:
I mispoke, this is obviously needed on the water module code. but in there we can easily use compiler macros to check for stat64 availability

falkTX:
forgot about this before, merging now, thanks!

PR #1807, opened 2023-09-05
JUCE: Fix build on musl

vimproved:
From https://github.com/juce-framework/JUCE/pull/1239.


vimproved:
Nevermind, I see that JUCE is not really used anymore. I'll close this.

PR #1808, opened 2023-09-05
JUCE: Fix build on musl

vimproved:
From https://github.com/juce-framework/JUCE/pull/1239.
Sorry for duplicate pull request, closed the original because I was slightly confused. This one should be correct.

### DIFF
--- a/source/modules/juce_core/juce_core.cpp
+++ b/source/modules/juce_core/juce_core.cpp
@@ -103,7 +103,7 @@
  #include <net/if.h>
  #include <sys/ioctl.h>
 
- #if ! (JUCE_ANDROID || JUCE_WASM)
+ #if ! (JUCE_ANDROID || JUCE_WASM || JUCE_MUSL)
   #include <execinfo.h>
  #endif
 #endif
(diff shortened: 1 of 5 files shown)


falkTX:
non mac/win builds do not use juce anymore, so this is not needed.

juce is moving too fast and does not care to maintain fixes for older releases, meaning a lot of maintenance work just to keep things up to date.
plus they now require C++17 while Carla is fine with C++11. do not want to bump the requirements just for the small things juce provides.

anyhow, this basically means juce is going away again. it is mostly a question of time.

what would be extremely useful is to add a CI step in the build.yml that tests builds using musl

vimproved:
> non mac/win builds do not use juce anymore, so this is not needed.
> 
> juce is moving too fast and does not care to maintain fixes for older releases, meaning a lot of maintenance work just to keep things up to date. plus they now require C++17 while Carla is fine with C++11. do not want to bump the requirements just for the small things juce provides.
> 
> anyhow, this basically means juce is going away again. it is mostly a question of time.
> 
> what would be extremely useful is to add a CI step in the build.yml that tests builds using musl

Agreed, a CI step for musl would be helpful. I know that you've been moving away from JUCE however juce_core.cpp is still being built (and failing without this patch) in 2.5.6 for me.

falkTX:
the 2.5.x series are in a separate branch, I still do releases based on 2.5 because the main/develop target is taking longer than expected.
the biggest blocker (add-plugin dialog redesign is 95% complete now though, so very soon releases will use the 2.6 series)

focusing efforts on 2.5.x for this repo is a waste of time. of course feel free to keep your patches for current stable release, but they wont be needed here.

Issue #1568, opened 2022-01-11
Add option to name Carla Patchbays


Bleuzen:
Hi, I want to use Carla-Patchbay as a plugin chain host for multiple "chains", so multiple instances are needed.
But when I start new Carla-Patchbays, there is currently no option to set the name, so I end up with this:

![image](https://user-images.githubusercontent.com/12885163/149029974-5d9417d6-6b0c-4d3d-b30b-1b8992e10f5e.png)

Which is confusing, not really clear which is my mic processing chain here.

Some way to name the patchbay instances would help here, at least for console.

I start them with:

```
carla-patchbay -n ~/Carla/mic-patchbay.carxp 
```

There could be a new command line option for example:

```
carla-patchbay -n --name "MicFX" ~/Carla/mic-patchbay.carxp 
```

falkTX:
why not simply use patchbay instances as plugins inside carla?
then you can just rename the plugins.

Bleuzen:
I want to load/unload any patchbay at any time, so each patchbay would need its own Carla instance, which means I end up with two Carla windows per fx chain, which is kinda ugly and can lead to worse overview.
Also, a command line option makes it more flexible. This way, Carla patchbays could easly be started/stopped from scripts with different names and bind to hotkeys.

falkTX:
Well I dont have that implemented now, but env var is. So this will work just the same:
```
CARLA_CLIENT_NAME=testing carla-patchbay
```

Daniel-I-Am:
I ran into a similar use case and noticed it only sometimes worked. I need predictable names for my inputs and outputs and would love the configuration ease of Carla.

I stumbled upon the env var before and when setting `CARLA_CLIENT_NAME` to some value, the pipewire inputs and outputs are properly registered with that name. However when starting it from a systemd unit file it no longer seems to work.

Some further debugging got me to this:  
Running Carla-rack without a GUI `--no-gui` breaks this functionality.

Poking around in code leads me to believe it is set here, but this code is not called when the GUI is not rendered.

https://github.com/falkTX/Carla/blob/ecad5268e71faa9d9d8a5d40b7573a6936ad4b6e/source/frontend/carla_host.py#L212

Do you have any idea if there is a way to change the output names to be predictable when not using a GUI for Carla?

---

<details><summary>My minimal example</summary>
<p>

```
➜  ~ pw-link -i
Midi-Bridge:Midi Through:(playback_0) Midi Through Port-0
[...]

PR #1811, opened 2023-09-26
Allow configuration of Carla client name when headless

Daniel-I-Am:
I read this issue https://github.com/falkTX/Carla/issues/1568, because I was running into a similar issue. Now that I had some time to poke the code, I noticed a relatively simple solution to my problem.

This should allow modifying the name of the pipewire/jack device when running headless mode with the same configuration option as when running with UI mode enabled.

Setting `CARLA_CLIENT_NAME` to any value will make the devices pick up that name instead of the default `"Carla"`.


falkTX:
simple and effective, will merge, thanks!

PR #1898, opened 2024-06-25
fix TCP unregister systematically fails

Houston4444:
Hi.
The OSC TCP `/unregister` method systematically fails, because code compares the url argument to the saved hostname.

This method is used by `carla-control`, it means that it is not possible to stop carla-control, restart it and register again to the same carla instance.

This PR fixes the problem of course.

I need it because I use a self made TCP server connected to Carla to save/restore my presets.


falkTX:
Fixed in 95e0482b898544c20b57d5c5ce0aee91514664a0 in a slightly different way.
Doing it based on the URL argument means any IP/host could send that message, it is better to do it the same way as for the registration, so it always ends up matching.

Issue #156, opened 2014-11-02
[feature request]  display LV2 plugin help


x42:
Port descriptions as tooltips in the patchbay would be nice.
And Plugin description (help text) for the plugin itself.

LV2 offers a simple plain-text (no markup) in `rdfs:comment` for both.

falkTX:
Carla's API has been extended to support comments for parameters, they are now used as tooltips in the "edit dialog".
Not sure if we can set tooltips on QGraphics stuff, need to investigate.

PR #1961, opened 2025-01-28
add-description-tab

jpka-:
Hello. I prepare this intermediate patch which uses Qt _setHtml_() as markup renderer, should use lv2 'rdfs:comment' field in **plaintext** or **html** form, and planned for **markdown**. It is not includes most tricky part, how to pass this rdfs:comment.
Thanks.
This is duplicate of https://github.com/falkTX/Carla/discussions/1960

### DIFF
--- a/source/frontend/carla_widgets.py
+++ b/source/frontend/carla_widgets.py
@@ -7,6 +7,11 @@
 
 from abc import abstractmethod
 
+import markdown
+
+import subprocess
+import shlex
+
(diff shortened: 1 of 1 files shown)


jpka-:
I've add further changes, not sure if it was done correctly within this pull, or should they be separated. There is also extra dependency of '**libgig**' now required, i provide also code with message if it's impossible. Thanks.

jpka-:
Related: #156

PR #1969, opened 2025-02-06
redraw-Qt-knobs

jpka-:
Three types of knobs (_internal, calf, openav_) redrawn with Qt only; no substrate svg/png image need now. Now hover enlighting works on all knobs. All is vectorized and use HSL color model. This is a complete work, intended to be replacement of current `scalabledial.py` file, tested also by CPU usage.
But it is one way ticket: png/svg stuff removed.


jpka-:
Correction:
- Added changes for very latest code;
- Removed warnings at log.

Issue #11, opened 2013-03-02
request: more options for knob control


wolftune:
I'd like a modifier key that switches the scroll-wheel on knobs and parameters between being very subtle or giving a bigger change. Right now, it is very subtle, moves very slowly. There could be a preference setting for default being slow small change vs larger quicker change. Then the modifier key could be used to get the alternate behavior.

There should also be a preference to let the cursor click-and-drag be linear even for the circular knobs.

rhetr:
is scrolling over knobs already implemented? I'm unable to get it to work for me.

jpka-:
Hello. I also thought the mouse scroll wheel not working, unless i use plugins with integer parameters (ladspa TAP reverb). So it works, but with quite incorrect speed. On floats it **extremely** slow, need about 3333 pulses per full knob revolution. Ctrl/Shift rotates it 3 times faster. Alt rotates it back. Contrary, on integers, it is too fast. 3 positions per wheel step, or 10 with modifier (so unable to select integer between 3's boundaries). Floats speed fixed to full revolution. Integer speed fixed to min-max values.
Dive into source codes, i see that all steps calculated correctly. But just not used in UI, which stuck to these turtle or cheetah speeds.
I can live with that, either make mouse drag from mouse wheel with bash script (drag works correctly); or, i can modify tap plugins to use integers and stretch min-maxes by three. Both ways are quite weird and it can be much more interesting if maintainer can fix the step value lost somewheere on the way to Qt UI (is looks like bug, not an unimplemented feature).
Thanks.

jpka-:
I also found that it is possible with 'qt5ct' to decrease wheel speed from 3x to 1x (for integers). It is under 'Interface -> Scroll lines'. Fast step remains to be 10x. While this is exactly as i need, but it is not so useful, because i still need 3x scroll for texts.
And it show that this magic 3x constant is not part of Carla.
Other Qt apps with knobs, like 'QSynth', are not depend of this setting.
In 'xev' i always see one 'press' per wheel pulse. 
P.S. I have 'QT_QPA_PLATFORMTHEME=qt5ct' in '/etc/environment', i use Archlinux.

jpka-:
[...]

PR #1970, opened 2025-02-10
create-Qt-pushbuttons

jpka-:
This commit supersedes my commit #1969.

1. Three types of **knobs** (internal, calf, openav) redrawn with Qt only; no substrate svg/png image need now. Now hover enlighting works on all knobs. All is vectorized and use HSL color model. This is a complete work, intended to be replacement of current scalabledial.py file, tested also by CPU usage.
2. But it is one way ticket: png/svg stuff removed.
3. Three types of 2- and 3-positions **pushbuttons** created. The idea is to make buttons well different from knobs, while still keep skin style, include disabled items. Used now for **booleans**, and **integers** @ [0,1], [0,1,2] ranges.
4. Mouse wheel captured for QDial, it solves #11 and gives exact control for bools/ints, while smooth one for floats. Use with _Ctrl_ or _Shift_ also.
5. Left mouse button toggles pushbuttons.
6. Hotkeys introduced when mouse hover active (no mouse press need) for both pushbuttons and knobs: _0...9_ to set absolute value; _Home_, _End,_ _PgUp,_ _PgDn._ Note: 'Home' interferes with Canvas control hotkey.
7. 
Hope it will be useful.

### DIFF
--- a/source/frontend/carla_skin.py
+++ b/source/frontend/carla_skin.py
@@ -6,6 +6,7 @@
 # Imports (Global)
 
 from qt_compat import qt_config
+import math
 
 if qt_config == 5:
     from PyQt5.QtCore import Qt, QRectF, QLineF, QTimer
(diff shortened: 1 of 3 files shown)


jpka-:
Please see updated #1971 which should include this code.

PR #1985, opened 2025-03-26
Fix broken program and bank messages for Jack applications.

kramlie:
It appears to simply be a typo. MAP_PROGRAM_CHANGES is not an available option for Jack applications, only SEND_PROGRAM_CHANGES is.


falkTX:
Yeah I think the "old" code was a copy&paste mistake, thanks for the correction!

PR #1987, opened 2025-04-24
Fix support for CYGWIN

carlo-bramini:
Carla doesn't work on CYGWIN and the message:
```
Unsupported platform!
```
is printed on the console when building.
The macro `CARLA_OS_UNIX` is needed to make a working executable for CYGWIN, so I added `CARLA_OS_CYGWIN`, to be included together with existing macros. Hopefully, it is an easy fix and it has solved all the problems here. Tested with latest development version of Lmms at the time of writing.

### DIFF
--- a/source/includes/CarlaDefines.h
+++ b/source/includes/CarlaDefines.h
@@ -38,6 +38,8 @@
 # define CARLA_OS_WIN32
 #elif defined(__APPLE__)
 # define CARLA_OS_MAC
+#elif defined(__CYGWIN__)
+# define CARLA_OS_CYGWIN
 #elif defined(__HAIKU__)
 # define CARLA_OS_HAIKU
 #elif defined(__linux__) || defined(__linux)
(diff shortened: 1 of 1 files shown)


falkTX:
cygwin is just windows, no?

is there a way to do CI build tests for this?

carlo-bramini:
> cygwin is just windows, no?

Actually, no.
CYGWIN is a POSIX environment that runs on top of Windows and it aims to provide the most faithful experience to the one you can get on UNIX-like operating systems. So, it has a dedicated libc library based on NEWLIB instead of the traditional MSVCRT or UCRT. Although the POSIX layer masks the existence of Windows, CYGWIN also provides a subset of W32API, that you can use for using Windows APIs while keeping use of POSIX functions. This must be handled with a bit of care, however.

 
> is there a way to do CI build tests for this?

Unfortunately, I have not experience of CI builds, I'm very sorry of that.

PR #2020, opened 2025-08-26
Add support for parameter labels also when running as a plugin.

kramlie:
Previously it was only available when running as a Jack application.


kramlie:
@falkTX: Anything else you need from me here?

kramlie:
Just a friendly reminder for this PR. I hope it can be considered for mainline.

falkTX:
of course! I dont have time to dedicated to carla for a little while, once I do I will handle this and a few other PRs all at once.

PR #2034, opened 2026-01-03
INSTALL.md: Add Qt6 support notes and Debian install instructions

krekhovx:


INSTALL.md
```diff
@@ -44,6 +44,7 @@ Optional for extended LV2 UIs support: (Linux only)
 
  - Qt4
  - Qt5
+ - Qt6
```
falkTX: this looks good, except this line.
there is no Qt6 LV2 UI support, as that is not a thing that exists.

falkTX:
nice, thanks!

PR #2040, opened 2026-03-19
Handle multiple buffer sizes in multi-client mode

hoshinolina:
In multiple clients mode, each plugin is a separate client and could have its own buffer size. In particular, with PipeWire, a buffer size change might be queued for the main client, and not delivered before the process callback is called for a plugin client, causing segfaults.

Hook up the buffer size / sample rate callback to clients, and do not call them from the engine code in multi client mode.

Note: By default, PipeWire synchronizes the buffer size of all JACK clients. However, this change is still necessary even then, because PipeWire does not guarantee delivery of a buffer size callback for one client before `process()` is called for another client. For example, this:

* (buffer size change 128 -> 256)
* plugin1: `bufsizechanged(256)`
* plugin1: `process(256)`
* mainengine: `bufsizechanged(256)`
* mainengine: `process(256)`

is possible, and this causes segfaults with Carla on PipeWire even with no extra configuration settings, unless the quantum is globally set in advance, since as plugin connections are made during startup the "global JACK buffer size" will change.

To actually test mixed buffer sizes in Carla on a continuous basis, you have to do something like this:

```
$ cat  .config/pipewire/jack.conf.d/test.conf 
jack.rules = [
    {
        matches = [
            {
                client.name = "Audio Gain (Stereo)"
            }
        ]
        actions = {
            update-props = {
                node.group = "mygroup"
            }
        }
    }
]
```

This will cause a plugin called `Audio Gain (Stereo)` to be scheduled with its own buffer size, not synchronized with all other JACK clients. Then, connecting that plugin to an audio device with one default buffer size, and (any) different plugin to another audio device with another default buffer size, will cause Carla to be processing a different buffer size for each plugin.

Fixes common segfaults during Carla initialization, or while changing buffer sizes at runtime.

### DIFF
--- a/source/backend/engine/CarlaEngine.cpp
+++ b/source/backend/engine/CarlaEngine.cpp
@@ -2249,14 +2249,19 @@ void CarlaEngine::bufferSizeChanged(const uint32_t newBufferSize)
 
     pData->time.updateAudioValues(newBufferSize, pData->sampleRate);
 
-    for (uint i=0; i < pData->curPluginCount; ++i)
+#ifndef BUILD_BRIDGE
+    if (pData->options.processMode != ENGINE_PROCESS_MODE_MULTIPLE_CLIENTS)
+#endif
     {
-        if (const CarlaPluginPtr plugin = pData->plugins[i].plugin)
+        for (uint i=0; i < pData->curPluginCount; ++i)
         {
-            if (plugin->isEnabled() && plugin->tryLock(true))
+            if (const CarlaPluginPtr plugin = pData->plugins[i].plugin)
             {
-                plugin->bufferSizeChanged(newBufferSize);
-                plugin->unlock();
+                if (plugin->isEnabled() && plugin->tryLock(true))
+                {
+                    plugin->bufferSizeChanged(newBufferSize);
+                    plugin->unlock();
+                }
             }
         }
     }
@@ -2278,14 +2283,19 @@ void CarlaEngine::sampleRateChanged(const double newSampleRate)
 
     pData->time.updateAudioValues(pData->bufferSize, newSampleRate);
 
-    for (uint i=0; i < pData->curPluginCount; ++i)
+#ifndef BUILD_BRIDGE
+    if (pData->options.processMode != ENGINE_PROCESS_MODE_MULTIPLE_CLIENTS)
+#endif
     {
-        if (const CarlaPluginPtr plugin = pData->plugins[i].plugin)
+        for (uint i=0; i < pData->curPluginCount; ++i)
         {
-            if (plugin->isEnabled() && plugin->tryLock(true))
+            if (const CarlaPluginPtr plugin = pData->plugins[i].plugin)
             {
-                plugin->sampleRateChanged(newSampleRate);
-                plugin->unlock();
+                if (plugin->isEnabled() && plugin->tryLock(true))
+                {
+                    plugin->sampleRateChanged(newSampleRate);
+                    plugin->unlock();
+                }
             }
         }
     }
--- a/source/backend/engine/CarlaEngineJack.cpp
+++ b/source/backend/engine/CarlaEngineJack.cpp
@@ -838,9 +838,27 @@ public:
         {
             CARLA_SAFE_ASSERT_RETURN(fJackClient != nullptr && ! isActive(),);
 
+#ifndef BUILD_BRIDGE
+            // When JACK is activated, some callbacks will be called synchronously.
+            // These callbacks must not re-lock the plugin, or else they would
+            // deadlock. Set a flag to tell them to elide locking, since we are
+            // already in a safe context (no other thread can lock the plugin).
+            fActivating = true;
+#endif
             try {
                 jackbridge_activate(fJackClient);
             } catch(...) {}
+#ifndef BUILD_BRIDGE
+            // Release the activation flag. From this point onwards non-RT callbacks
+            // will lock the plugin. Since one such callback could have been called
+            // twice, with the second instance not blocking after activate and
+            // concurrently executing, we lock the callback lock before doing this.
+            fCallbackLock.lock();
+            fActivating = false;
+            fCallbackLock.unlock();
+            // After this point, no callbacks will be running nor able to run until
+            // the plugin mutex is released in CarlaPlugin::setEnabled().
+#endif
         }
 
         CarlaEngineClient::activate();
@@ -1191,9 +1209,31 @@ public:
     }
 #endif
 
+#ifndef BUILD_BRIDGE
+    void maybeLockPlugin(CarlaPluginPtr plugin)
+    {
+        fCallbackLock.lock();
+        if (!fActivating)
+            plugin->tryLock(true);
+    }
+
+    void maybeUnlockPlugin(CarlaPluginPtr plugin)
+    {
+        if (!fActivating)
+            plugin->unlock();
+
+        fCallbackLock.unlock();
+    }
+#endif
+
 private:
     jack_client_t* fJackClient;
     const bool     fUseClient;
+
+#ifndef BUILD_BRIDGE
+    CarlaMutex     fCallbackLock;
+    bool           fActivating;
+#endif
 
     LinkedList<CarlaEngineJackAudioPort*> fAudioPorts;
     LinkedList<CarlaEngineJackCVPort*>    fCVPorts;
@@ -2005,10 +2045,8 @@ public:
 
 #ifndef BUILD_BRIDGE
             pluginReserve = new CarlaPluginPtr(plugin);
-            /*
-            jackbridge_set_buffer_size_callback(fClient, carla_jack_bufsize_callback_plugin, pluginReserve);
-            jackbridge_set_sample_rate_callback(fClient, carla_jack_srate_callback_plugin, pluginReserve);
-            */
+            jackbridge_set_buffer_size_callback(client, carla_jack_bufsize_callback_plugin, pluginReserve);
+            jackbridge_set_sample_rate_callback(client, carla_jack_srate_callback_plugin, pluginReserve);
             // jackbridge_set_latency_callback(client, carla_jack_latency_callback_plugin, pluginReserve);
             jackbridge_set_process_callback(client, carla_jack_process_callback_plugin, pluginReserve);
             jackbridge_on_shutdown(client, carla_jack_shutdown_callback_plugin, pluginReserve);
@@ -4421,25 +4459,39 @@ private:
         return 0;
     }
 
-    /*
     static int JACKBRIDGE_API carla_jack_bufsize_callback_plugin(jack_nframes_t nframes, void* arg)
     {
-        CarlaPlugin* const plugin((CarlaPlugin*)arg);
-        CARLA_SAFE_ASSERT_RETURN(plugin != nullptr && plugin->isEnabled(), 0);
+        CarlaPluginPtr* const pluginPtr = static_cast<CarlaPluginPtr*>(arg);
+        CARLA_SAFE_ASSERT_RETURN(pluginPtr != nullptr, 0);
 
+        CarlaPluginPtr plugin = *pluginPtr;
+        CARLA_SAFE_ASSERT_RETURN(plugin.get() != nullptr && plugin->isEnabled(), 0);
+
+        CarlaEngineJackClient* const engineClient((CarlaEngineJackClient*)plugin->getEngineClient());
+        CARLA_SAFE_ASSERT_RETURN(engineClient != nullptr, 0);
+
+        engineClient->maybeLockPlugin(plugin);
         plugin->bufferSizeChanged(nframes);
+        engineClient->maybeUnlockPlugin(plugin);
         return 1;
     }
 
     static int JACKBRIDGE_API carla_jack_srate_callback_plugin(jack_nframes_t nframes, void* arg)
     {
-        CarlaPlugin* const plugin((CarlaPlugin*)arg);
-        CARLA_SAFE_ASSERT_RETURN(plugin != nullptr && plugin->isEnabled(), 0);
+        CarlaPluginPtr* const pluginPtr = static_cast<CarlaPluginPtr*>(arg);
+        CARLA_SAFE_ASSERT_RETURN(pluginPtr != nullptr, 0);
 
+        CarlaPluginPtr plugin = *pluginPtr;
+        CARLA_SAFE_ASSERT_RETURN(plugin.get() != nullptr && plugin->isEnabled(), 0);
+
+        CarlaEngineJackClient* const engineClient((CarlaEngineJackClient*)plugin->getEngineClient());
+        CARLA_SAFE_ASSERT_RETURN(engineClient != nullptr, 0);
+
+        engineClient->maybeLockPlugin(plugin);
         plugin->sampleRateChanged(nframes);
+        engineClient->maybeUnlockPlugin(plugin);
         return 1;
     }
-    */
 
     static void JACKBRIDGE_API carla_jack_latency_callback_plugin(jack_latency_callback_mode_t /*mode*/, void* /*arg*/)
     {


falkTX:
I don't like this idea... pipewire wants to have a replacement for jack, but this breaks how jack works (both jack1 and jack2).

Your proposal shifts `bufferSizeChanged` from the "offline" side into audio/realtime processing, which sounds terrible to me. Buffer size changes are not RT safe, calling it during the process function is going to lead to a lot of sudden xruns. Granted it happens after some buffer size change so there are always audio discontinuities, but this clearly breaks RT behaviour which is the main blocker for me.

Better that pipewire be fixed rather than every application to be fixed instead.
I would consider the pipewire behaviour here to be the broken one, not Carla.

Assuming a global sample rate and buffer size has always been a guarantee in the jack API.

hoshinolina:
What you are asking for is in fact realtime-breaking. The crash is happening *because* PipeWire can handle buffer change notifications without causing global xruns.

To emulate behavior that does not crash Carla, PipeWire would have to, on every buffer size change, suspend *all* processing of *every* JACK client, call all the buffer size changed notifications, *wait for those callbacks to return*, then resume processing. This means that a single laggy client in that buffer size callback would block every single JACK client. In other words, you want that callback to be in the realtime hot path for every single JACK client.

Without doing that, there is no way to emulate behavior that would cause the current version of Carla not to crash, because it relies on the global engine buffer size callback getting called and completing *before* any plugin engines process any new cycles.

Ultimately, Carla here is relying on a JACK bug/misfeature, that a single client can block the entire JACK server and every single client. Without implementing that anti-feature, this cannot be fixed on the JACK side. It is not reasonable for Carla to have multiple independent JACK clients and expect the callbacks to have any ordering in the way they're called between them. If they are independent clients, they are independent clients. This only worked on JACK "by accident".

PipeWire *does* implement a global sample rate and buffer size for JACK clients by default, but that is not enough here. You are asking for buffer size changes to be a globally blocking event with no timeout. That is not reasonable.

What this patch does is move the buffer realloc to the *first* RT cycle after the buffer size change. This is not, in fact, a regression, because given the above, for JACK to work at all, it has to behave like that already (the buffer size change callbacks have to block the process cycles). And this is entirely reasonable if you think about it, because changing the buffer size means allocating memory, which is not RT-safe, so no matter what you do, there *will* be RT-unsafe code when the buffer size changes.

To put another way: `process()` cannot run with the new buffer size until the bufsize callback completes, even for a single client. Therefore, bufsize is already on the RT hotpath, and always was.

hoshinolina:
It would be possible to do the buffer size change directly in the callback. This would still block the process cycle for that client anyway (it has to), so it wouldn't actually improve things in any significant way, but if you insist:

The reason I moved it to `process()` is that that callback cannot unconditionally take the plugin lock, because it is called during the activation process, which has the lock held, so it would deadlock. To fix this, the plugin locking would have to be reworked so that the plugin lock is not held during the JACK activation call (and probably also deactivation and others).

falkTX:
> To emulate behavior that does not crash Carla, PipeWire would have to, on every buffer size change, suspend _all_ processing of _every_ JACK client, call all the buffer size changed notifications, _wait for those callbacks to return_, then resume processing. This means that a single laggy client in that buffer size callback would block every single JACK client. In other words, you want that callback to be in the realtime hot path for every single JACK client.

well yes, because this is how JACK works.

falkTX:
> It is not reasonable for Carla to have multiple independent JACK clients and expect the callbacks to have any ordering in the way they're called between them. If they are independent clients, they are independent clients. This only worked on JACK "by accident".

not by accident, it is a core principle of jack that there is only 1 "master" device. and changing buffer size is a global operation that applies to all clients, as everything is in sync.

hoshinolina:
And PipeWire does keep the same quantum for all JACK clients by default (unless you do an override like I did). What it does not is *synchronously* call the callbacks across every single JACK client.

I don't think it's reasonable to assume/expect that multiple independent JACK clients will just block each other like that. Sure JACK used to do that, and that was a *bad* thing since it caused a single misbehaving client to xrun or hang the entire world.

falkTX:
> What this patch does is move the buffer realloc to the _first_ RT cycle after the buffer size change. This is not, in fact, a regression, because given the above, for JACK to work at all, it has to behave like that already (the buffer size change callbacks have to block the process cycles). And this is entirely reasonable if you think about it, because changing the buffer size means allocating memory, which is not RT-safe, so no matter what you do, there _will_ be RT-unsafe code when the buffer size changes.

while this is true, it is also a question of semantics and well defined rules:
- audio is allowed to stop for buffer size changes, and reallocations are allowed for that. (changing the buffer size of an audio device is not guaranteed to be a lock-free operation)
- audio processing should be RT safe

both can be true and apply to how jack works already today

hoshinolina:
OK, so then I'll just figure out how to avoid the deadlock in the non-RT path then...

falkTX:
> The reason I moved it to `process()` is that that callback cannot unconditionally take the plugin lock, because it is called during the activation process, which has the lock held, so it would deadlock. To fix this, the plugin locking would have to be reworked so that the plugin lock is not held during the JACK activation call (and probably also deactivation and others).

are you sure it deadlocks? if process and buffer size callbacks are from the same thread, they won't be called concurrently. if they are separate threads there should be no dead-lock, the process tryLock will fail if the other thread is busy, and buffer size callbacks will take the lock that waits for processing to end.
am I missing something here...?

hoshinolina:
Yes, you end up with this:

```
Thread 30 (Thread 0x7fff919ef6c0 (LWP 657986) "data-loop.0"):
#0  __futex_lock_pi64 (futex_word=futex_word@entry=0x555556ac9038, clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0) at futex-internal.c:180
#1  0x00007ffff76826de in __pthread_mutex_lock_full (mutex=0x555556ac9038) at pthread_mutex_lock.c:441
#2  0x00007fffc3534e1c in CarlaMutex::lock (this=0x555556ac9038) at ../../utils/CarlaMutex.hpp:76
#3  0x00007fffc35bb443 in CarlaBackend::CarlaPlugin::tryLock (this=0x555556accaf0, forcedOffline=true) at CarlaPlugin.cpp:2323
#4  0x00007fffc35a88e8 in CarlaBackend::CarlaEngineJack::carla_jack_bufsize_callback_plugin (nframes=1024, arg=0x555556ae9750) at CarlaEngineJack.cpp:4461
#5  0x00007fffd40e5bd3 in do_emit_buffer_size (loop=<optimized out>, async=<optimized out>, seq=<optimized out>, data=<optimized out>, size=<optimized out>, user_data=<optimized out>) at ../pipewire-jack/src/pipewire-jack.c:4850
#6  0x00007fffdc01087d in flush_all_queues (impl=0x555556ada1a8) at ../spa/plugins/support/loop.c:366
#7  0x00007fffdc00d7f4 in source_event_func (source=0x555556ab8de0) at ../spa/plugins/support/loop.c:893
#8  0x00007fffdc00f506 in loop_iterate_cancel (object=0x555556ada1a8, timeout=<optimized out>) at ../spa/plugins/support/loop.c:684
#9  0x00007fffd4030d51 in do_loop (user_data=0x555556ada0a0) at ../src/pipewire/data-loop.c:67
#10 0x00007ffff767f464 in start_thread (arg=<optimized out>) at pthread_create.c:448
#11 0x00007ffff77025ec in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78


Thread 1 (Thread 0x7ffff7e9abc0 (LWP 657851) "carla"):
#0  __syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56
#1  0x00007ffff767bc3c in __internal_syscall_cancel (a1=<optimized out>, a2=<optimized out>, a3=a3@entry=8, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=0) at cancellation.c:49
#2  0x00007ffff767bc84 in __syscall_cancel (a1=<optimized out>, a2=<optimized out>, a3=a3@entry=8, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=0) at cancellation.c:75
#3  0x00007ffff76f573e in __GI___libc_read (fd=<optimized out>, buf=<optimized out>, nbytes=nbytes@entry=8) at ../sysdeps/unix/sysv/linux/read.c:26
#4  0x00007fffdc022097 in read (__fd=<optimized out>, __buf=<optimized out>, __nbytes=8) at /usr/include/bits/unistd.h:32
#5  impl_eventfd_read (object=<optimized out>, fd=<optimized out>, count=<optimized out>) at ../spa/plugins/support/system.c:216
#6  0x00007fffdc011062 in spa_system_eventfd_read (object=<optimized out>, fd=<optimized out>, count=0x7fffffffb670) at ../spa/include/spa/support/system.h:194
#7  loop_queue_invoke (object=<optimized out>, func=<optimized out>, seq=<optimized out>, data=<optimized out>, size=<optimized out>, block=<optimized out>, user_data=<optimized out>) at ../spa/plugins/support/loop.c:477
#8  loop_invoke (object=<optimized out>, func=<optimized out>, seq=<optimized out>, data=<optimized out>, size=<optimized out>, block=<optimized out>, user_data=<optimized out>) at ../spa/plugins/support/loop.c:545
#9  0x00007fffd40eee2e in jack_activate (client=0x555556ad4d10) at ../pipewire-jack/src/pipewire-jack.c:4896
#10 0x00007fffc3675759 in jackbridge_activate (client=0x555556ad4d10) at JackBridge1.cpp:961
#11 0x00007fffc359c123 in CarlaBackend::CarlaEngineJackClient::activate (this=0x555556a39340) at CarlaEngineJack.cpp:842
#12 0x00007fffc35b76c8 in CarlaBackend::CarlaPlugin::setEnabled (this=0x555556accaf0, yesNo=true) at CarlaPlugin.cpp:1385
#13 0x00007fffc354f05d in CarlaBackend::CarlaEngine::loadProjectInternal (this=0x55555642c060, xmlDoc=..., alwaysLoadConnections=false) at CarlaEngine.cpp:3140
#14 0x00007fffc3547a5d in CarlaBackend::CarlaEngine::loadProject (this=0x55555642c060, filename=0x7fffc1f47950 "/home/lina/bufsize-test.carxp", setAsCurrentProject=true) at CarlaEngine.cpp:1420
```

In JACK2 it works like this:

* JackClient::Activate()
  * StartThread()
    * JackPosixThread::StartSync()
      * Set status to kStarting
      * StartImp() and poll for !kStarting
         * Thread: JackPosixThread::ThreadHandler()
            * Set status kIniting
            * Init()
              * fBufferSize(...) **calls the buffer size callback from the realtime thread** (!)
            * Set status kRunning

There is a comment that says:
```
    /*
        Execute buffer_size callback.

        Since StartThread uses fThread.StartSync, we are sure that buffer_size callback
        is executed before StartThread returns (and then IsActive will be true).
        So no RT callback can be called at the same time.
    */
```

However, the status is set to kIniting *before* running Init(), and that causes StartThread to return early.

So in fact, JACK2 works **because it is broken**, and it has a race condition where the bufsize callback might be called reentrantly from the realtime thread (the first time) and then asynchronously from another thread on notifications.

If that race condition were fixed, JACK2 would also deadlock.

The issue is not the process callback, it's that the first buffer size callback is (or at least intended to be) made synchronously when `jack_activate()` is called.

hoshinolina:
Pushed a version that does not touch the process callback. This instead uses a separate lock and flag to avoid double-locking the plugin during the activation process, in a safe and controlled manner.

falkTX:
Thank you, the new `<atomic>` ended up not being used so can be removed right?

So this would mean the deadlock is due to `activate()` calling context already having the plugin lock taken, which would re-lock again due to jack client activation waiting for the callbacks to complete.

I think it is the right direction, but leads me to think about deeper underlying issues in Carla if this was not caught before.

hoshinolina:
Yeah, already removed the atomic (forgot that from an experiment).

The whole thing with "async" callbacks was always kind of a bit weird yeah, since as I said they are RT-blocking anyway kind of by definition (at least the ones that affect the audio path). That's why PipeWire itself does generally process everything in a single thread (that is realtime for audio cases). The deadlock happens anyway due to a cross-thread dependency though.

This isn't the first instace of PW not being "100% compatible" with JACK and crashing some apps, but the JACK API was always kind of not very well designed... same reason JACK2 used to crash some apps that JACK1 didn't until they got fixed etc.

(FWIW Carla has always been "crashy on startup" for me for years on PipeWire, and also on explicit buffer size changes (happens to other people too). I believe this is the bug that was causing all that. I'm going to recommend Carla for complex audio processing setups on Linux for streaming/etc, so I wanted to get this one fixed before it ends up leaving a bad impression due to the flakiness...)

I did test outright different buffer sizes as I mentioned in the original post and it does work, which is kind of cool if you think about it. It means with multi client mode and the right poking in the `jack.conf`, you can have multiple disconnected graphs within the same Carla instance running with different quantums and latencies. I think if you use dummy loopback devices (that behave as disjoint nodes) you might even be able to have audio flow across quantum size boundaries in principle...

Here's how it looks in pw-top:

```
S   ID  QUANT   RATE    WAIT    BUSY   W/Q   B/Q  ERR FORMAT           NAME
I   29      0      0   0.0us   0.0us  ???   ???     0                  Dummy-Driver
S   30      0      0    ---     ---   ---   ---     0                  Freewheel-Driver
I  179      0      0   0.0us   0.0us  ???   ???     0                  Midi-Bridge
R  134    256  48000  62.6us   1.2us  0.01  0.00   19   S32LE 32 48000 alsa_input.usb-RME_Digiface_USB__24184482__000080F082DB888-00.pro-input-0
R  138    256      0   8.3us   9.6us  0.00  0.00    0   S32LE 34 48000  + alsa_output.usb-RME_Digiface_USB__24184482__000080F082DB888-00.pro-output-0
R  329      0      0  23.4us   0.3us  0.00  0.00   34                   + Carla
R   52      0      0  26.5us  11.3us  0.00  0.00    0                   + Audio Gain (Stereo) (2)
R  283   1024  48000  49.6us  22.5us  0.00  0.00    0    S32LE 2 48000 alsa_output.pci-0000_05_00.1.hdmi-stereo-extra2
R  169      0      0  10.7us  30.9us  0.00  0.00    0                   + Audio Gain (Stereo)
```

falkTX:
> This isn't the first instace of PW not being "100% compatible" with JACK and crashing some apps, but the JACK API was always kind of not very well designed... same reason JACK2 used to crash some apps that JACK1 didn't until they got fixed etc.

not sure if it is the API but more of the inner workings of the implementation.
changing buffer size in jack is not a common action, there are even some famous clients that stop processing audio altogether after the first buffer size change (ahem zita stuff ahem)

I am glad someone is paying attention to this, I mostly run jack on my systems so my experience with pipewire is limited.

anyhow I can't find the comment I left before, but my proposal is to do something like: (I didnt check this code just an idea)

```
Mutex& getCallbackLock(bool& isActivating) const noexcept
{
    isActivating = fIsActivating;
    return fCallbackLock;
}
```

I think it looks a bit cleaner.

source/backend/engine/CarlaEngine.cpp
```diff
@@ -2278,14 +2280,16 @@ void CarlaEngine::sampleRateChanged(const double newSampleRate)
 
     pData->time.updateAudioValues(pData->bufferSize, newSampleRate);
 
-    for (uint i=0; i < pData->curPluginCount; ++i)
-    {
-        if (const CarlaPluginPtr plugin = pData->plugins[i].plugin)
+    if (pData->options.processMode != ENGINE_PROCESS_MODE_MULTIPLE_CLIENTS) {
```
falkTX: please put the last { in newline, as per the code style in Carla

falkTX:
Another thing - I think this breaks standalone bridges running without carla (the `BUILD_BRIDGE` macro) because they use a single global client and don't register buffersize callbacks.
So the new vars should be contained inside a `#ifndef BUILD_BRIDGE` and also the check of "if (processMode != multi-client)" too.

hoshinolina:
> anyhow I can't find the comment I left before, but my proposal is to do something like: (I didnt check this code just an idea)
> 
> ```
> Mutex& getCallbackLock(bool& isActivating) const noexcept
> {
>     isActivating = fIsActivating;
>     return fCallbackLock;
> }
> ```
> 
> I think it looks a bit cleaner.

Well that's not going to work because the lock needs to be locked before reading fIsActivating... and returning the guard does not work either because it's not copyable (nor moveable).

falkTX:
it doesnt need to copy or move, it returns a reference directly.

My suggestion only applies to the jack callbacks, not for the initial assignment of the fIsActivating just before the jack_activate()

hoshinolina:
The order of operations has to be:

- Lock the callback lock
- Read fActivated
- Call plugin function (locking if necessary)
- Unlock the callback lock

Reading fActivated outside the lock is undefined behavior.

falkTX:
right yes, sorry.

hoshinolina:
Made the other style/ifdef changes, not sure what to do about the functions then...

falkTX:
> Made the other style/ifdef changes, not sure what to do about the functions then...

Its ok, I am trying to apply better code practices in Carla but the codebase is a bit old and for sure not what I would consider a good example :relaxed: 
So assuming the CI builds still pass with the latest change, I will merge it.
Thanks!

Then need to backport into the 2.5.x hotfix branch, as the main branch has some WIP details and is not good state for release.

hoshinolina:
@falkTX Apparently this deadlocks with VST3 plugins on project load: https://bugzilla.redhat.com/show_bug.cgi?id=2457554

I don't understand why though. The relevant part of the backtrace is:

```
#12 0x00007fffbd033a4e in juce::MessageManager::Lock::tryAcquire(bool) const [clone .constprop.0] (this=this@entry=0x7fff837fd998, lockIsMandatory=false) at ../modules/juce_events/messages/juce_MessageManager.cpp:351
#13 0x00007fffbcff1f9c in juce::MessageManager::Lock::tryEnter (this=0x7fff837fd998) at ../modules/juce_events/messages/juce_MessageManager.cpp:309
#14 juce::MessageManagerLock::attemptLock (threadToCheck=0x0, jobToCheck=0x0, this=0x7fff837fd990) at ../modules/juce_events/messages/juce_MessageManager.cpp:432
#15 juce::MessageManagerLock::MessageManagerLock (threadToCheck=0x0, this=0x7fff837fd990) at ../modules/juce_events/messages/juce_MessageManager.cpp:411
#16 juce::VST3PluginInstance::prepareToPlay (this=this@entry=0x5555559b59b0, newSampleRate=48000, estimatedSamplesPerBlock=estimatedSamplesPerBlock@entry=2048) at ../modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp:2502
#17 0x00007fffbcf11019 in CarlaBackend::CarlaPluginJuce::activate (this=<optimized out>) at plugin/CarlaPluginJuce.cpp:1016
#18 CarlaBackend::CarlaPluginJuce::bufferSizeChanged (newBufferSize=<optimized out>, this=<optimized out>) at plugin/CarlaPluginJuce.cpp:1517
#19 CarlaBackend::CarlaPluginJuce::bufferSizeChanged (this=<optimized out>, newBufferSize=<optimized out>) at plugin/CarlaPluginJuce.cpp:1507
#20 0x00007fffbcea344f in CarlaBackend::CarlaEngineJack::carla_jack_bufsize_callback_plugin (nframes=2048, arg=<optimized out>) at engine/CarlaEngineJack.cpp:4482
```

So JUCE is trying to acquire the MessageManagerLock, but I don't understand how that could already be locked by another thread.

hoshinolina:
OK, I think I understand better. It's not being called from the message manager thread, so it posts a message to try to block the message manager thread, but that thread isn't running to process it.

~~I also see Carla has a non-JUCE VST3 implementation, so I told the user to try `CARLA_DO_NOT_USE_JUCE_FOR_VST3`. Is that production ready in the 2.5 branch? As far as I can tell JUCE is gone in main already...~~

So the non-JUCE VST3 support in 2.5 is incomplete, I see. I'm not sure what to do here, is there some way to fix the JUCE message thread thing? If not, I guess no backport to 2.5 and we'll have to revert in Fedora...

PR #2071, opened 2026-07-28
CarlaEngineJack: allow MIDI events over 254 bytes on input

luginf:
EngineMidiEvent::size was uint8_t, and the JACK input paths (both the per-port event class and the Rack/Patchbay ingestion code) asserted size < 0xFF before truncating-casting to uint8_t, silently dropping any incoming SysEx bigger than 254 bytes (e.g. a full patch/timbre bulk dump, as opposed to short parameter-sized SysEx which already worked).

Widened EngineMidiEvent::size and fillFromMidiData's size parameter to uint16_t, and fixed both occurrences of the assert+cast in CarlaEngineJack.cpp accordingly.

Scoped to direct/in-process JACK hosting (Rack, Patchbay, Multiple Clients); out-of-process bridged plugins still serialize MIDI event size as a single byte over IPC and would need separate follow-up work


luginf:
**Known remaining limitation**

This fix only covers plugins hosted directly (Rack/Patchbay/Multiple Clients modes). If a plugin runs through Carla's separate-process "bridge" (out-of-process VST/LV2 hosting, or the JACK-application bridge), the same message-size limit can still silently drop very large SysEx dumps there. That's because the message size is also hard-coded as a single byte in a couple of internal communication protocols between Carla and the bridged process, which would each need to be extended (both the sending and receiving side, kept in sync) to lift the limit - a notably bigger change than this build's fix, and not something most users are likely to hit in a typical direct-hosting setup like this one.

PR #2077, opened 2026-08-15
water: bound the MIDI file header parse (malformed .mid crashes MidiFile::readFrom)

manuaudio:
A malformed `.mid` crashes `MidiFile::readFrom`. Reachable from a user-chosen file: the **MIDI File** native plugin (`source/native-plugins/midi-file.cpp`, also shipped as an LV2 bundle) calls `readFrom` on whatever is loaded.

## The bug

`MidiFileHelpers::parseMidiHeader` takes the MThd chunk length straight from the file and never checks it:

```cpp
unsigned int bytesRemaining = ByteOrder::bigEndianInt (data);
...
bytesRemaining -= 6;
data += bytesRemaining;
```

A declared length below six wraps the subtraction; any declared length at all advances the read pointer that far past the buffer. The caller then computes `size -= (size_t) (d - base)`, which wraps in turn, so the track-chunk loop believes it has an enormous amount of data left and reads from wherever the pointer landed.

The `RIFF` branch is a second, independent path: it reads eight 4-byte words looking for `MThd` with no size check whatsoever, so a 20-byte file overruns before the header is even located.

## Reproduced under ASan

Five files, none larger than 26 bytes:

| file | result |
|---|---|
| MThd length `0` | `BUS` in `ByteOrder::bigEndianInt` |
| MThd length `5` | `BUS` |
| MThd length `0xFFFFFFFF` | `BUS` |
| MThd length `1000` | `heap-buffer-overflow` |
| `"RIFF"` + 16 zero bytes | `heap-buffer-overflow` |

Built with `-fsanitize=address,undefined` against `source/modules/water/water.cpp` and a small driver that feeds the bytes to `MidiFile::readFrom` via a `MemoryInputStream`.

The only existing guard is `size > 16` in `readFrom`, which all five files satisfy.

## The change

Thread the remaining byte count through `parseMidiHeader`, check it before every read, refuse a header chunk length that is under six or longer than the file, and clamp the track-chunk loop to what is actually left rather than trusting the declared `chunkSize`.

## Evidence it doesn't break valid files

Built before and after (binaries hash-checked as different), then compared parses:

| file | before | after |
|---|---|---|
| format 0, 1 track | `tracks=1 timeFormat=96` | identical |
| format 1, 3 tracks | `tracks=3 timeFormat=96` | identical |
| SMPTE division `0xE728` | `tracks=1 timeFormat=-6360` | identical |
| MThd length 10 (legal longer header) | `tracks=1 timeFormat=96` | identical |
| RIFF-wrapped | `tracks=1 timeFormat=96` | identical |

All five malformed files are now rejected with `readFrom` returning false and no sanitizer report.

## Two things I looked at and left alone

- `convertTicksToSeconds` divides by `(timeFormat & 0xff)` and by `(timeFormat & 0x7fff)`, both of which a malformed header can zero. `time` is a `double`, so these are floating-point divisions yielding `inf` rather than undefined behaviour. It produces nonsense timestamps rather than a crash, so I have left it out of a fix aimed at the crash — happy to add a guard if you want one.
- `ByteOrder::bigEndianInt` does unaligned 4-byte loads, which UBSan reports on *valid* files too. Pre-existing and unrelated to this change; I mention it only so it isn't mistaken for fallout.

`water` looks to be a fork of an older JUCE; upstream may well have tightened this since, so there may be a tidier patch to lift. This one is deliberately minimal and local.


falkTX:
Any kind of LLM based submissions are not welcome or wanted in Carla, thank you but no thank you.

